PR #3, opened 2018-08-20
Add missing part of LICENSE

sfendrich:
The recommended way of using the LGPL3 license is to include both
the GPL3 and the LGPL3 text (https://www.gnu.org/licenses/gpl-howto.en.html).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,165 +1,674 @@
-                   GNU LESSER GENERAL PUBLIC LICENSE
+                    GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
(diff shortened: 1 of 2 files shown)


sfendrich:
The main point (missing part) was that the LGPL3-text contains only the exceptions from the GPL3 and is incomplete without the main GPL3-text. Concerning the other points:
- Both  COPYING and LICENSE are widely used, e.g., COPYING is the standard in GNU projects, while LICENSE is standard in Apache-projects. Therefore, I didn't care which variant we use and chose the one that was already there. 
- I vote for adding a license notice in each file

PR #7, opened 2018-08-29
jdbc backend: stream osh cells for the given cell id range

tyrasd:
The old implementation always loads all osh cells for the given cell id range into in a static collection before streaming them out (to the parallel stream that invokes the CellIterator, etc.).

The new implementation makes loads the data as needed by converting the jdbc `ResultSet` into an `Iterable` and streams that (as outlined here: https://dzone.com/articles/adding-java-8-lambda-goodness). :sunglasses:


tyrasd:
Forgot to post the benchmark results:

```
# before

building histocounts: 2.93s ±0.25s
highway histocounts (length): 3.21s ±0.17s
forest histocounts (area): 7.67s ±0.21s

# after

building histocounts: 2.70s ±0.16s
highway histocounts (length): 2.68s ±0.16s
forest histocounts (area): 7.14s ±0.22s
```

Overall this even results in a slight performance improvement.

PR #9, opened 2018-08-31
applying naming and link changes

FabiKo117:
OSHDB should be written in upper-case
removed links to internal confluence pages
removed links to internal gitlab pages
updated links to public github pages
changed name of keytablesOshdb variable to keytables


tyrasd:
Thanks Fabian! :bouquet:

PR #10, opened 2018-08-31
Faster getModificationTimestamps routines

### LINKED ISSUE #8: Endless loading in OSHRelation.getModificationTimestamps()
When sending a simple request over the ohsome-API about bicycle routes for a small bbox in Rheinland-Pfalz (Germany), the code seems to be somehow stuck (or needs far too much time) within the OSHRelation.getModificationTimestamps() method. The reason seems to be that the relation https://www.openstreetmap.org/relation/88376#map=9/49.9658/7.3759 is quite big and has lots of versions and members with loads of different timestamps.
<p>
This is the request URI: /elements/count?time=2012-06-15&showMetadata=true&keys=route&values=bicycle&bboxes=7.6563,49.4635,7.7727,49.513&types=relation

tyrasd:
Improves the way modification timestamps for OSHEntity objects are calculated:

* avoid nested loops – this did work OK for small entities like ways with a couple of versions, but not for large (route) relations that have thousands of members, have hundreds of versions and it's members also have dozens of versions each. Before complexity was ~`O(v*m*mv)`, now it should be more like `O(v+m*mv*f)`, where `f` is a small factor (larger than one) representing the amount of fragmentation of the membership time intervals of the particular relation (which is usually small).
* calculate filtered modification timestamps directly (without first calculating all modification timestamps and filtering them afterwards).
* test osh entities early with the given osm entity filter (e.g. when filtering for `highway=residential`, ways that have the `highway` tag but never had the value `residential` don't need to be looked at in further processing steps).

This resolves #8

### DIFF
--- a/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/celliterator/CellIterator.java
+++ b/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/celliterator/CellIterator.java
@@ -1,5 +1,6 @@
 package org.heigit.bigspatialdata.oshdb.util.celliterator;
 
+import com.google.common.collect.Streams;
 import com.vividsolutions.jts.geom.*;
 import java.io.Serializable;
 import java.util.*;
@@ -160,6 +161,10 @@ public class CellIterator implements Serializable {
       )) {
         // this osh entity doesn't match the prefilter or is fully outside the requested
         // area of interest -> skip it
+        continue;
+      }
+      if (Streams.stream(oshEntity).noneMatch(osmEntityFilter)) {
+        // none of this osh entity's versions matches the filter -> skip it
         continue;
       }
       boolean fullyInside = allFullyInside || (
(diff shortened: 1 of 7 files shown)


tyrasd:
Here's a quick performance test:

```
    OSHDBH2 oshdb = new OSHDBH2("…/rheinland-pfalz.oshdb");
    Integer result = OSMEntitySnapshotView
        .on(oshdb)
        .areaOfInterest(new OSHDBBoundingBox(7.458,49.016,8.236,49.58))
        .timestamps("2018-01-01")
        .osmType(OSMType.RELATION)
        .osmTag("type", "route")
        .osmEntityFilter(e -> e.getId() == 88376)
        .count();
```

This takes ~ 8 min on current master (and oshdb v0.4), and ~5 sec on this branch.

The normal benchmarks also seem to be slightly faster:

```
# PR

building histocounts: 0.96s ±0.41s
highway histocounts (length): 0.73s ±0.01s
forest histocounts (area): 5.60s ±0.05s
unique contributors (simple area of interest): 1.29s ±0.11s
count roads (complex area of interest): 3.50s ±0.06s
object history by id: 0.23s ±0.02s

# master

building histocounts: 1.28s ±0.55s
highway histocounts (length): 0.84s ±0.02s
forest histocounts (area): 7.03s ±0.66s
unique contributors (simple area of interest): 1.28s ±0.07s
count roads (complex area of interest): 3.67s ±0.10s
object history by id: 0.31s ±0.01s
```

Issue #17, opened 2018-09-06
stream results


tyrasd:
Sometimes, one just wants to stream non-aggregated data (e.g. to write it out into a file of DB), where the current workaround (of `.collect()`-ing all results into a temporary list and then streaming it out) is not perfect because it unnecessarily requires quite a lot of memory. There could be a `.stream()` method in the `MapReducer` which could return (chunks of) results as soon as they are ready.

Maybe this could also expanded to the `MapAggregator` class, ~~where instead of a single stream one would get a set of streams~~?!

PR #19, opened 2018-09-18
Implement MapReducer.stream()

tyrasd:
Implements the idea from #17 

Works in JDBC (singlethread/multithread) backends as well as the "AffinityCall" ignite backend (other backends fall back to `.collect().stream()`. For the other ignite backends it is not very straight-forward to implement. Maybe it can be implemented at some later stage, or the oshdb can (by default) fall back to another backend if it finds that the current one doesn't implement streaming?

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/Kernels.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/Kernels.java
@@ -3,9 +3,11 @@ package org.heigit.bigspatialdata.oshdb.api.mapreducer.backend;
 import com.google.common.collect.Iterables;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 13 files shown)


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/Kernels.java
```diff
@@ -0,0 +1,266 @@
...
+  static <R, S> CellProcessor<S> getOSMContributionGroupingCellReducer(
+      SerializableFunction<List<OSMContribution>, Iterable<R>> mapper,
+      SerializableSupplier<S> identitySupplier,
+      SerializableBiFunction<S, R, S> accumulator,
+      CancelableProcessStatus aborter
+  ) {
+    return (oshEntityCell, cellIterator) -> {
+      AtomicReference<S> accInternal = new AtomicReference<>(identitySupplier.get());
+      // iterate over the history of all OSM objects in the current cell
+      List<OSMContribution> contributions = new ArrayList<>();
```
tyrasd: btw: while testing, I found that here (and in a few places below) re-using this array for multiple consecutive entities can cause a problem, when one just directly uses these arrays as output, e.g. the following would produce different results:

```java
OSMContributionView.….groupByEntity().collect() // <- wrong result
OSMContributionView.….groupByEntity().map(list -> new ArrayList(list)).collect()
```

This is kind of a corner case, since (at first glance) it only happens when one does not specify and map function and does not aggregate the results. But it could still be confusing and (I think) it doesn't matter too much from a performance point of view. I guess we should just not re-use the list here, or what do you think?

PR #30, opened 2018-10-04
ignite backend: add onClose callback

tyrasd:
This can be used to close connections to (temporary) databases that were used to store or retrieve intermediate data in a query. This PR also expose a few internal states (e.g. `TagTranslator.getConnection()`, `LazyEvaluatedObject.wasEvaluated()`) to be accessed from external applications.

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/db/OSHDBIgnite.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/db/OSHDBIgnite.java
@@ -12,6 +12,7 @@ import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.ignite.Ignite;
 import org.apache.ignite.Ignition;
+import org.apache.ignite.lang.IgniteRunnable;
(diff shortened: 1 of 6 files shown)


tyrasd:
Here is a schematic example usage where one auxiliary database connection is opened (and properly closed) on each ignite worker.

helper class:

```java
public class RemoteObject<T> extends LazyEvaluatedObject<T> {
  public RemoteObject<T>(SerializableSupplier<T> evaluator) {
    super(evaluator);
  }

  @Override
  public synchronized T get() {
    return super.get();
  }
}
```

use with oshdb query:

```java

RemoteObject<Connection> remote = new RemoteObject<>(() -> {
  Class.forName("…");
  return DriverManager.getConnection("…", "user", "passwd");
});

OSHDBIgnite oshdbIgnite = new OSHDBIgnite("…").onClose(() -> {
  if (remote.wasEvaluated()) {
    remote.close();
  }
});

// run an oshdb query and access remote db connection
…
  .…
  .map(x -> {
    Connection local = remote.get();
    // now one can execute an sql query using the ignite worker's local db connection
    …
    return something;
  })
  .…
```

PR #33, opened 2018-10-10
improve performance of FastPolygonOperations constructor

tyrasd:
Replaces the straight-forward `O(n²)` polygon slicing implementation with a quad-recursive `O(n log(n))` algorithm (assuming that JTS' intersection method takes O(n) time to compute the intersection of a n-vertex polygon with another 4 vertex polygon). A minor additional change is that now, the number of bands is rounded up to the next power of two (mainly to make the implementation easier).


tyrasd:
Some quick & dirty benchmarks (timings are for a full oshdb-api request including oshdb processing time, ignite data serialization and other oshdb-api overhead; old code executed directly on the cluster, new code was run from a regular desktop machine):

| region  | old implementation | this pull request |
| ------------- | ------------- | ------------- |
| ~10k Points<br>(Thüringen boundary) | 1.9 s | 3.5 s |
| ~60k Points Multipolygon<br>(Germany boundary) | 24 s | 151 s |

PR #34, opened 2018-10-11
quantiles

tyrasd:
This uses the [t-digest](https://raw.githubusercontent.com/tdunning/t-digest/master/docs/t-digest-paper/histo.pdf) method (and code) by Ted Dunning and Otmar Ertl to estimate quantiles of the distribution of the result set. A short description of the method can be [found here](https://dataorigami.net/blogs/napkin-folding/19055451-percentile-and-quantile-estimation-of-big-data-the-t-digest).

Todo:
* [x] tweak t-disgest parameters (`compression`)?
* [x] implement for `MapAggregator`
* [x] unit tests

### DIFF
--- a/oshdb-api/pom.xml
+++ b/oshdb-api/pom.xml
@@ -90,6 +90,12 @@
       <artifactId>annotations</artifactId>
       <version>${jetbrainsannotations.version}</version>
     </dependency>
+
+    <dependency>
+      <groupId>com.tdunning</groupId>
+      <artifactId>t-digest</artifactId>
+      <version>3.2</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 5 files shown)


tyrasd:
This feature is actually quite cool. I did a quick analysis of the distribution of the length of osm highways:

highway-length-quantiles

It's also quite interesting to see that the road lenghts almost perfectly follow a [log-normal distribution](https://en.wikipedia.org/wiki/Log_normal_distribution) (as can be seen in this [QQ-plot](https://en.wikipedia.org/wiki/Q%E2%80%93Q_plot)):

highway-length-qqplot

tyrasd:
@rabidllama suggested to name these methods more precisely (i.e. reflect that they don't return the _exact_ quantiles or median, but only a statistical estimation). Which is actually a good point, as it avoids potential confusions of users that maybe otherwise would expect precise results.

Maybe `.estimatedQuantile` could work… what do you think?

sfendrich:
Yes, avoiding this confusion is important. `.estimatedQuantile` is fine.

tyrasd:
@sfendrich 

> Rename methods such as quantile to estimatedQuantile or something similar in order to avoid wrong expectations.

done in 741b6c880be133d4798476cdeaf3daea9faeda91

PR #35, opened 2018-10-18
changing name of 'changeset' to 'changesetId'

FabiKo117:
adding getChangesetId()
marking getChangeset() as deprecated


oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/osm/OSMEntity.java
```diff
@@ -52,8 +52,16 @@ public OSHDBTimestamp getTimestamp() {
     return timestamp;
   }
 
+  public long getChangesetId() {
+	    return changesetId;
+	  }
```
tyrasd: looks like there are some superfluous tabulators above

```suggestion
    return changesetId;
  }
```

PR #50, opened 2019-01-14
fix calculation of fullyInside: test against oshEntity bbox

tyrasd:
The existing solution only uses the optimized path for the case: _"is the cell fully completely inside the query bounding box"_ when working on a bounding polygon query, because the check `bboxInPolygon.test(boundingBox)` is always false (because `boundingBox` is the extent of the polygon used in the `bboxInPolygon` check). It is however possible and intended to test whether _"the current entity is fully inside the query polygon"_ to further optimize such queries.


tyrasd:
> was the null check unnecessary?

Yeah, that's a left-over from the very early times where the GeometryBuilder returned `null` for stuff it couldn't construct geometries for (e.g., unsupported relation types, ways with 0 nodes, etc.). Now for these cases, "proper" geometries are returned (e.g., a geometrycollection objects, or empty geometries).

PR #54, opened 2019-01-15
Patch z grid

beikerta:


rtroilo:
This patch is not relevant any more, we can/should drop it.

PR #56, opened 2019-01-16
less memory intensive implementation of localKeys getter

tyrasd:
Especially with large queries that span over a lot of cells, the existing implementation (storing every cell key in a long list) is not very memory efficient. Using a stream for this should be much more efficient.

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteLocalPeek.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteLocalPeek.java
@@ -5,6 +5,9 @@ import com.vividsolutions.jts.geom.Polygonal;
 import java.io.Serializable;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.ignite.Ignite;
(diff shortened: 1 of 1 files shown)


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteLocalPeek.java
```diff
@@ -225,33 +228,30 @@ public boolean isActive() {
...
+                  .map(key -> new ImmutablePair<>(cache, key));
+            });
       });
-      Collections.shuffle(localKeys);
-      return localKeys;
     }
 
     S execute(Ignite node, CellProcessor<S> cellProcessor) {
-      return this.localKeys(node).parallelStream()
+      return this.localKeys(node).parallel()
```
rtroilo: I doubt that .parallel() has any effect at this point, cellIdRange in its current implementation is not parallelizable, and so the rest of the stream.
Suggestion is to get an Iterator  `this.localKeys(node).iterator()` then batch up some localKeys and reduce them in parallel.
```java
ArrayList<Object> batch = new ArrayList<>(parallelism);
while(itr.hasNext()){
	batch.add(itr.next());
	if(batch.size() == parallelism){
		batch.stream()
		.parallel();
		// map&filter
		// reduce batch
		batch.clear();
	}
}
batch.stream()
.parallel();
// map&filter
// reduce batch
batch.clear();
//reduce batch results together
```

tyrasd: > I doubt that .parallel() has any effect at this point

Are you sure? If I try something like this, it seems like `.parallel` can be called on a stream generated from `StreamSupport.stream(…, false)`:

```java
Iterable<Integer> foo = () -> (new ArrayList<Integer>()).iterator();
System.out.println(StreamSupport.stream(foo.spliterator(), false).isParallel());
System.out.println(StreamSupport.stream(foo.spliterator(), false).parallel().isParallel());
```

This returns `false`, then `true` (`isParallel` returns whether the stream, […] would execute in parallel).

:thinking:

rtroilo:
```java
final int limit = 1_000;

Iterable<Long> test = () -> new Iterator<Long>(){
	private long l = 0;
	@Override
	public boolean hasNext() {
		return l < limit;
	}

	@Override
	public Long next() {
		if(!hasNext())
			throw new NoSuchElementException();
		return Long.valueOf(l++);
	}
};

Consumer<Long> burn10ms = it -> {
	try {
		Thread.sleep(10);
	} catch (InterruptedException e) {
	}
};

Stopwatch stopwatch = Stopwatch.createStarted();
StreamSupport.stream(test.spliterator(), true).parallel()
.forEach(burn10ms);
System.out.println(""+stopwatch);

ArrayList<Long> test2 = new ArrayList<>(limit);
for(int i=0; i< limit; i++){
	test2.add(Long.valueOf(i));
}
stopwatch.reset().start();
StreamSupport.stream(test2.spliterator(), false).parallel()
.forEach(burn10ms);
System.out.println(""+stopwatch);
```
According to this test, it would not be parallelized.

tyrasd: > I'm not sure if isParallel return if the stream is really parallel.

According to the javadoc it should: https://docs.oracle.com/javase/8/docs/api/java/util/stream/BaseStream.html#isParallel--

> According to this test

When I try it with a stream with more than only 1,000 elements, it seems to run in parallel just fine (I replaced the `Thread.sleep(10)` with a loop that burs a "few" CPU cycles, and can get all my CPUs working at close to 100%).

PR #64, opened 2019-01-25
improve parallelization in localPeek backend

tyrasd:
Sadly, #56 had some negative side effects regarding performance: Because the top-most stream in the pipeline was a super short one (i.e. max. 3 entries for the different osm types), the java stream implementation never parallelizes it, even though the `flatMap` step adds many many entries to the stream (at least in OpenJDK 8 – see https://stackoverflow.com/a/25706806/1627467). Also, the way the `cellIds` were distributed to the different threads was not optimal: A single thread was likely to receive a large portion of the typically large low zoom level cells (that's why the initial implementation had the
[`Collections.shuffle(localKeys);`](https://github.com/GIScience/oshdb/pull/56/files/d4a62091a5ae53a754cfc91c368b042c94c4b6a5#diff-2f3781c256730a1206a3368a569b94f6L249) in there).

Here's an improved solution (based on the idea by @rtroilo in the [original PR](https://github.com/GIScience/oshdb/pull/56#discussion_r249383013)):

* the stream is now based on an iterator which uses an internal buffer of 1 million cell ids
* these are shuffeled to ensure that no thread has too many large cells to work through
* then distributed accross threads using a parallel java stream
* for now, it also removes the initial [`igniteNode.affinity.mapKeysToNodes`](https://github.com/GIScience/oshdb/pull/56/files/d4a62091a5ae53a754cfc91c368b042c94c4b6a5#diff-2f3781c256730a1206a3368a569b94f6L244) "filtering" step (and just perform `localPeek` on all potential cell ids anyway), because it doesn't seem to make much of a difference (if anything, queries are a bit slower because of the increased amount of sequential code) and just adds unnecessary complexity to the code.

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteLocalPeek.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteLocalPeek.java
@@ -241,14 +241,14 @@ class IgniteLocalPeekHelper {
       return this.notCanceled;
     }
 
-    private class CellIdIterator implements Iterator<CellId> {
+    private class CellKeysIterator implements Iterator<Long> {
       Iterator<Pair<CellId, CellId>> cellIdRanges;
-      ArrayList<CellId> buffer;
+      ArrayList<Long> buffer;
 
       // a buffer of about 1M interleaved cellIds
       static final int BUFFER_SIZE = 1_000_000;
 
-      CellIdIterator(Iterable<Pair<CellId, CellId>> cellIdRanges) {
+      CellKeysIterator(Iterable<Pair<CellId, CellId>> cellIdRanges) {
         this.cellIdRanges = cellIdRanges.iterator();
         buffer = new ArrayList<>(BUFFER_SIZE);
       }
@@ -264,7 +264,7 @@ class IgniteLocalPeekHelper {
           long fromId = cellIdRange.getLeft().getId();
           long toId = cellIdRange.getRight().getId();
           for (long id = fromId; id < toId; id++) {
-            buffer.add(new CellId(level, id));
+            buffer.add(new CellId(level, id).getLevelId());
           }
         }
         Collections.shuffle(buffer);
@@ -282,7 +282,7 @@ class IgniteLocalPeekHelper {
       }
 
       @Override
-      public CellId next() {
+      public Long next() {
         return buffer.remove(buffer.size() - 1);
       }
     }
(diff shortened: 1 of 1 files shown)


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteLocalPeek.java
```diff
@@ -226,34 +241,73 @@ public boolean isActive() {
       return this.notCanceled;
     }
 
-    public abstract S execute(Ignite node);
+    private class CellKeysIterator implements Iterator<Long> {
+      Iterator<Pair<CellId, CellId>> cellIdRanges;
+      ArrayList<Long> buffer;
+
+      // a buffer of about 1M interleaved cellIds
+      static final int BUFFER_SIZE = 1_000_000;
```
rtroilo: Better to you a buffer-size with a multiple of 1024, because it seems 1024 this is the buffer size for parallel streams.

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteLocalPeek.java
```diff
@@ -226,34 +241,73 @@ public boolean isActive() {
...
+          if (!cellIdRanges.hasNext()) {
+            break;
+          }
+          Pair<CellId, CellId> cellIdRange = cellIdRanges.next();
+          int level = cellIdRange.getLeft().getZoomLevel();
+          long fromId = cellIdRange.getLeft().getId();
+          long toId = cellIdRange.getRight().getId();
+          for (long id = fromId; id < toId; id++) {
+            buffer.add(new CellId(level, id).getLevelId());
+          }
```
rtroilo: we could put this code 
```java
          Pair<CellId, CellId> cellIdRange = cellIdRanges.next();
          int level = cellIdRange.getLeft().getZoomLevel();
          long fromId = cellIdRange.getLeft().getId();
          long toId = cellIdRange.getRight().getId();
          for (long id = fromId; id < toId; id++) {
            buffer.add(new CellId(level, id).getLevelId());
          }
```
in a flatMap step of the cellIdRanges stream. that makes pulling out BUFFER_SIZE ids more predictable.
At the moment it could happen that we "overfill" the buffer with more than BUFFER_SIZE ids.

tyrasd: does https://github.com/GIScience/oshdb/pull/64/commits/0f50172a43a14a7f48c166934c1bb44441c150cf#diff-2f3781c256730a1206a3368a569b94f6R255 look ok?

PR #65, opened 2019-01-29
maven clean up, fix checkstyle reporting

SlowMo24:
messing with Maven Dependency mess. upgrading some dependencies, removing unused reporting (maven site), enabling personal checkstyle sheet

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,7 +52,7 @@ pipeline {
         }
       }
     }
-    
+
     stage ('Trigger Benchmark') {
       when {
         expression {
@@ -68,7 +68,7 @@ pipeline {
         }
       }
     }
-        
+
     stage ('Publish Javadoc'){
       when {
         expression {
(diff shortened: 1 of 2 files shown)


SlowMo24:
@rtroilo might be consulted first because maven is always tricky and hi knows

SlowMo24:
also I am sorry this commit comes AFTER the codestyle week. But it was only now requested by @tyrasd  :-)

pom.xml
```diff
@@ -71,6 +79,54 @@
...
+            <version>${pmd.version}</version>
+          </dependency>
+          <dependency>
+            <groupId>net.sourceforge.pmd</groupId>
+            <artifactId>pmd-java</artifactId>
+            <version>${pmd.version}</version>
+          </dependency>
+          <dependency>
+            <groupId>net.sourceforge.pmd</groupId>
+            <artifactId>pmd-javascript</artifactId>
```
tyrasd: do we need the javascript module as well?

Jenkinsfile
```diff
@@ -132,7 +126,7 @@ pipeline {
           
           //warnings plugin
           rtMaven.run pom: 'pom.xml', goals: '--batch-mode -V -e checkstyle:checkstyle pmd:pmd pmd:cpd findbugs:findbugs com.github.spotbugs:spotbugs-maven-plugin:3.1.7:spotbugs -Dmaven.repo.local=.m2'
-
+
```
tyrasd: normally, we try to avoid unnecessary white space like here…

pom.xml
SlowMo24: ok, your right, should have checked before copy paste... sorry

Jenkinsfile
SlowMo24: can't explain where they come from. thanks for noticing. Will catch up on this...

PR #67, opened 2019-01-30
avoid clipping if not necessary

tyrasd:
This applies the idea from https://github.com/GIScience/oshdb/pull/66 to the cellIterator functions: Some objects might have a large OSH bbox, but small (snapshot or contribution) geometries. Then it is not always necessary to calculate actual intersections, which are potentially slow computations.


tyrasd:
I've rebased this onto #111, since it has similar goal (increased performance of geometry generation). Let's follow up there.

PR #71, opened 2019-02-11
More robust building of polygon geometries from bboxes

SlowMo24:
OSHDBGeometryBuilder was returning empty Geometry for a point-like Bounding-Box. Now we force rectangle-like polygons.

### DIFF
--- a/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
+++ b/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
@@ -229,29 +229,29 @@ public class OSHDBGeometryBuilder {
         boolean joinable = false;
         for (int i = 0; i < ways.size(); i++) {
           List<OSMNode> what = ways.get(i);
-          if (lastId == what.get(0).getId()) { // end of partial ring matches to start of current
-                                               // line
+          if (lastId == what.get(0).getId()) {
+            // end of partial ring matches to start of current line
             what.remove(0);
             current.addAll(what);
             ways.remove(i);
             joinable = true;
             break;
-          } else if (firstId == what.get(what.size() - 1).getId()) { // start of partial ring
-                                                                     // matches end of current line
+          } else if (firstId == what.get(what.size() - 1).getId()) {
+            // start of partial ring matches end of current line
             what.remove(what.size() - 1);
             current.addAll(0, what);
             ways.remove(i);
             joinable = true;
             break;
-          } else if (lastId == what.get(what.size() - 1).getId()) { // end of partial ring matches
-                                                                    // end of current line
+          } else if (lastId == what.get(what.size() - 1).getId()) {
+            // end of partial ring matches end of current line
             what.remove(what.size() - 1);
             current.addAll(Lists.reverse(what));
             ways.remove(i);
             joinable = true;
             break;
-          } else if (firstId == what.get(0).getId()) { // start of partial ring matches start of
-                                                       // current line
+          } else if (firstId == what.get(0).getId()) {
+            // start of partial ring matches start of current line
             what.remove(0);
             current.addAll(0, Lists.reverse(what));
             ways.remove(i);
@@ -282,12 +282,16 @@ public class OSHDBGeometryBuilder {
   }
   
   /**
-   * Create the Geometry of an OSHDBBoundingBox. Will return a polygon with exactly 4 vertices even
-   * for point or line-like BoundingBox. Nevertheless the result might not pass the 
-   * {@link com.vividsolutions.jts.geom.Geometry#isRectangle() Geometry.isRectangle} test.
+   * Converts a OSHDBBoundingBox to a rectangular polygon.
+   *
+   * <p>
+   * Will return a polygon with exactly 4 vertices even for point or line-like BoundingBox.
+   * Nevertheless, for degenerate bounding boxes (width and/or height of 0) the result might
+   * not pass the {@link Geometry#isRectangle() Geometry.isRectangle} test.
+   * </p>
    *
    * @param bbox The BoundingBox the polygon should be created for.
-   * @return com.vividsolutions.jts.geom.Geometry Returns a Polygon for convenience.
+   * @return a rectangular Polygon
    */
   public static Polygon getGeometry(OSHDBBoundingBox bbox) {
     assert bbox != null : "a bounding box is not allowed to be null";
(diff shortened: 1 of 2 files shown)


oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -279,24 +279,31 @@ private static Geometry getMultiPolygonGeometry(
       T entity, OSHDBTimestamp timestamp, TagInterpreter areaDecider, P clipPoly) {
     Geometry geom = OSHDBGeometryBuilder.getGeometry(entity, timestamp, areaDecider);
     return Geo.clip(geom, clipPoly);
+  }
```
tyrasd: it looks like there is a superfluous space after the `}` here.

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -279,24 +279,31 @@ private static Geometry getMultiPolygonGeometry(
...
+   * Create the Geometry of an OSHDBBoundingBox. Will always contain 5 Coordinates, even for
+   * point-like bbx -> point-like polygons for bbx of a Point.
+   *
+   * @param bbox The BoundingBox the polygon should be created for.
+   * @return com.vividsolutions.jts.geom.Geometry Returns JTS geometry object for convenience.
+   * Result is an empty polygon for a null BBX.
+   */
+  public static Polygon getGeometry(OSHDBBoundingBox bbox) {
+    GeometryFactory gf = new GeometryFactory();
+    if (bbox == null) {
```
tyrasd: why the extra check against `null` here? As far as I know, we never generate null bounding boxes.
Maybe add an assert instead (`assert bbox != null : "a bounding box is not allowed to be null";` or so)?

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -279,24 +279,31 @@ private static Geometry getMultiPolygonGeometry(
...
     return Geo.clip(geom, clipPoly);
+  } 
+  
+  /**
+   * Create the Geometry of an OSHDBBoundingBox. Will always contain 5 Coordinates, even for
+   * point-like bbx -> point-like polygons for bbx of a Point.
+   *
+   * @param bbox The BoundingBox the polygon should be created for.
+   * @return com.vividsolutions.jts.geom.Geometry Returns JTS geometry object for convenience.
+   * Result is an empty polygon for a null BBX.
```
tyrasd: i think this should be indented according to our style guide. also, I think we never use the abbreviation _BBX_ for bounding boxes, but rather use _bbox_ if we want to have something short…

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -279,24 +279,31 @@ private static Geometry getMultiPolygonGeometry(
       T entity, OSHDBTimestamp timestamp, TagInterpreter areaDecider, P clipPoly) {
     Geometry geom = OSHDBGeometryBuilder.getGeometry(entity, timestamp, areaDecider);
     return Geo.clip(geom, clipPoly);
+  } 
+  
+  /**
+   * Create the Geometry of an OSHDBBoundingBox. Will always contain 5 Coordinates, even for
+   * point-like bbx -> point-like polygons for bbx of a Point.
```
tyrasd: I think we could write that the result is always a rectangular _polygon_ with exactly 4 vertices. (5 coordinate geometries could be many things :sweat_smile: )

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilderTest.java
```diff
@@ -160,6 +162,28 @@ public void testRelationGetGeometryIncomplete() {
     assertEquals("GeometryCollection", result.getGeometryType());
     assertEquals(1, result.getNumGeometries());
   }
+  
+  @Test
+  public void testBoundingBoxGetGeometry() {
+    OSHDBBoundingBox bbx = new OSHDBBoundingBox(0, 0, 0, 0);
```
tyrasd: maybe better start the test with the "regular" case, and then test the more specific corner cases like point/line degenerated bbox? also a comment line would be nice here to explain why these need to be tested differently.

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
tyrasd: oh, and is there a reason to mention the class of the result object here in the `@return` javadoc? I mean, doesn't javadoc see that from the definition of the function itself already?

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
SlowMo24: sorry for all the whytespace errors the last time. I need to get my IDE running with this...

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
SlowMo24: yes I thought of that too. Will implement...

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
SlowMo24: just copied the javadoc that was in the header before. Was wondering, too :-)

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
SlowMo24: well... the check for geometry.isRectangular() does not always return true (eg. if the bbox has the form of a line). But one of the first checks isRect.. does is _does it have exactly 5 coordinates) so I used this more generic description

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilderTest.java
SlowMo24: code-quality is really high on the agenda these days. I like it :+1: !!! Many things to learn for me.

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
SlowMo24: clarified in javadoc

PR #73, opened 2019-02-12
[Ignite] Late deserialize when off-heap store is used

tyrasd:
This delays loading of cache values, if they are stored off heap. This boosts performance for non global queries significantly (_scanquery2_ is the code from this PR, _scanquery_ is current master):

scanquery-performance

Interestingly, this also seems to reduce pressure on heap memory usage during global queries by a bit, which is a nice additional benefit.

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
@@ -488,10 +488,9 @@ public abstract class MapReducer<X> implements
       ret.filters.add(ignored -> false);
       return ret;
     }
-    int keyId = keyValueId.getKey();
-    int valueId = keyValueId.getValue();
+    OSHDBTagKey keyId = new OSHDBTagKey(keyValueId.getKey());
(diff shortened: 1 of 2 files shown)


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -488,10 +488,9 @@ protected MapReducer(MapReducer<?> obj) {
       ret.filters.add(ignored -> false);
       return ret;
     }
-    int keyId = keyValueId.getKey();
-    int valueId = keyValueId.getValue();
+    OSHDBTagKey keyId = new OSHDBTagKey(keyValueId.getKey());
     ret.preFilters.add(oshEntitiy -> oshEntitiy.hasTagKey(keyId));
-    ret.filters.add(osmEntity -> osmEntity.hasTagValue(keyId, valueId));
+    ret.filters.add(osmEntity -> osmEntity.hasTagValue(keyValueId.getKey(), keyValueId.getValue()));
```
tyrasd: Sorry, I accidentally committed this, since it isn't actually directly related to this PR. But it makes the methods (`osmTag(key)` and `osmTag(key, value)`) work more similar in the way they use the `preFilter` function.

PR #77, opened 2019-02-22
Syntax issue (fix proposal) on paragraph #03

kant:


tyrasd:
Btw: what's the point of `#03` in the commit message (and pull request title)? As far as I can see, this file contains only one paragraph. Did you perhaps find this typo by using some kind of automated process? Also, using the `#` character might generate unwanted links in the github user interface.

PR #79, opened 2019-02-26
cleaning dependencies, remove apache-commons

rtroilo:
I cleaned up the dependency hierarchy and removed the dependency to apache-commons.


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/generic/OSHDBCombinedIndex.java
```diff
@@ -12,16 +12,22 @@ public OSHDBCombinedIndex(U index1, V index2) {
...
 
   public V getSecondIndex() {
-    return this.payload.getRight();
+    return this.index2;
   }
 
   @Override
   public int compareTo(@NotNull OSHDBCombinedIndex<U,V> o) {
-    return this.payload.compareTo(o.payload);
+	final Comparable<Object> ci1 = (Comparable<Object>) index1;
```
tyrasd: there seems to be a tab character here. also below (line 27, 28)

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/generic/OSHDBCombinedIndex.java
```diff
@@ -12,16 +12,22 @@ public OSHDBCombinedIndex(U index1, V index2) {
...
+    int c = ci1.compareTo(o.index1);
+    if(c == 0) {
```
tyrasd: missing space between `if` and the `(`: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/GeometrySplitter.java
```diff
@@ -56,13 +54,33 @@
     this.subregions = subregions;
   }
 
+  public static class IndexData<I,D> {
+	  public final I index;
```
tyrasd: please also convert the tabs here to spaces

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/GeometrySplitter.java
```diff
@@ -56,13 +54,33 @@
...
+	  public final D data;
```
tyrasd: shouldn't this field be private?

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/GeometrySplitter.java
```diff
@@ -56,13 +54,33 @@
...
+	  }
+
+	public I getIndex() {
+		return index;
+	}
+
+	public D getData() {
+		return data;
+	}
+
```
tyrasd: I would reduce the amount of newline whitespace here

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/GeometrySplitter.java
```diff
@@ -108,7 +126,7 @@
           }
         }).collect(Collectors.toCollection(LinkedList::new));
   }
-
+
```
tyrasd: seems like there are some superfluous spaces here

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -997,4 +997,68 @@ public X setValue(X value) {
         TreeMap::new
     ));
   }
+  
+  /*
+   * A generic Pair class for holding key/value pairs
+   */
+  public static class Pair<U,V> {
+  	private U key;
```
tyrasd: tab -> spaces

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -997,4 +997,68 @@ public X setValue(X value) {
         TreeMap::new
     ));
   }
+  
+  /*
```
tyrasd: should this be a javadoc comment?

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -997,4 +997,68 @@ public X setValue(X value) {
...
+  	
+
+
+  	public U getKey() {
+  		return key;
+  	}
+
+
+
+  	public void setKey(U key) {
```
tyrasd: I think we don't need a setter for the key property.

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/GeometrySplitter.java
```diff
@@ -123,7 +141,7 @@
    * @param data the OSMContribution to split into the given sub-regions
    * @return a list of OSMContribution objects
    */
-  public List<Pair<U, OSMContribution>> splitOSMContribution(OSMContribution data) {
+  public List<IndexData<U,OSMContribution>> splitOSMContribution(OSMContribution data) {
```
tyrasd: I _think_, we could implement this whole thing as a `Map<U, OSMContribution>` instead of a `List<IndexData<U,OSMContribution>>` (since the list will at most contain one entry per index value). :thinking:

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -997,4 +997,68 @@ public X setValue(X value) {
...
+  public static class Pair<U,V> {
+  	private U key;
+  	private V value;
+  	
+  	public Pair(U key, V value) {
+  		this.key = key;
+  		this.value = value;
+  	}
+  	
+
```
tyrasd: reduce newlines maybe?

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/GeometrySplitter.java
tyrasd: then we could get rid of the `IndexData` class altogether. In the MapReducer class, we would use this method then like this:

```java
    ret = this.flatMap(x -> gs.splitOSMContribution((OSMContribution) x).entrySet())
            .aggregateBy(Entry::getKey, geometries.keySet()).map(Entry::getValue);
```

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -1790,7 +1790,7 @@ protected TagTranslator getTagTranslator() {
   }
 
   // get all cell ids covered by the current area of interest's bounding box
-  protected Iterable<Pair<CellId, CellId>> getCellIdRanges() {
+  protected Iterable<CellIdRange> getCellIdRanges() {
```
tyrasd: :+1:

oshdb-tool/etl/pom.xml
```diff
@@ -47,6 +47,12 @@
 			<artifactId>RoaringBitmap</artifactId>
 			<version>0.6.59</version>
 		</dependency>
+		
+		<dependency>
```
tyrasd: maybe we should now convert the whole file to use spaces instead of tabs?

oshdb-util/pom.xml
```diff
@@ -14,23 +14,30 @@
       <artifactId>oshdb</artifactId>
       <version>${project.version}</version>
     </dependency>
-    
+
```
tyrasd: extra spaces here

oshdb-util/pom.xml
```diff
@@ -14,23 +14,30 @@
...
     <dependency>
-      <groupId>com.h2database</groupId>
-      <artifactId>h2</artifactId>
-      <version>${h2.version}</version>
+      <groupId>com.googlecode.json-simple</groupId>
+      <artifactId>json-simple</artifactId>
+      <version>${googlejson.version}</version>
     </dependency>
 
+	<!-- TEST dependencies -->
```
tyrasd: tab

oshdb/pom.xml
```diff
@@ -19,72 +19,37 @@
...
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${guava.version}</version>
     </dependency>
     
-    <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-lang3</artifactId>
-      <version>${apachecommons.version}</version>
+	<!-- TEST dependencies -->
```
tyrasd: tab -> spaces

oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
```diff
@@ -47,7 +46,7 @@
   
   private static final long serialVersionUID = 1L;
   private static final Logger LOG = LoggerFactory.getLogger(XYGrid.class);
-  private static final double EPSILON = OSHDB.GEOM_PRECISION;
+//  private static final double EPSILON = OSHDB.GEOM_PRECISION;
```
tyrasd: maybe it would be better to remove it, if it's not used/needed

oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
```diff
@@ -197,6 +196,62 @@ public long getEstimatedIdCount(final OSHDBBoundingBox data) {
   public int getLevel() {
     return zoom;
   }
+  
+  public static class IdRange implements Comparable<IdRange>, Serializable{
```
tyrasd: put it into own source file? it seems to me that this is not specific to the XYGrid class
//edit: same for CellIdRange class below (https://github.com/GIScience/oshdb/pull/79/files#diff-d4b1ff67fed6306c1592fa6fdbbd9692R162)

tyrasd: also, tabs -> spaces

oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/util/OSHDBFilter.java
```diff
@@ -0,0 +1,24 @@
+package org.heigit.bigspatialdata.oshdb.util;
+
+import java.util.List;
+
+public interface OSHDBFilter {
+	
+	/**
```
tyrasd: tabs -> spaces

oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/util/byteArray/FastByteArrayOutputStream.java
```diff
@@ -0,0 +1,71 @@
+package org.heigit.bigspatialdata.oshdb.util.byteArray;
```
tyrasd: was this adapted/copied over from [`it.unimi.dsi.fastutil.io.FastByteArrayOutputStream`](https://github.com/vigna/fastutil/blob/master/src/it/unimi/dsi/fastutil/io/FastByteArrayOutputStream.java)? if yes, we should/must state that somewhere.

tyrasd: but don't we have this library as a dependency anyway? what's the use for this class here?

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -997,4 +997,68 @@ public X setValue(X value) {
         TreeMap::new
     ));
   }
+  
+  /*
+   * A generic Pair class for holding key/value pairs
+   */
+  public static class Pair<U,V> {
```
tyrasd: shouldn't this be in a separate java source file? also, I really liked the separation between mutable and immutable pairs in apache.commons.lang3. Would be nice if we kept that.

rtroilo: actually, I would be nice if this has a more meaning full name what describes for what this pair class is used for.
To make this class a nested class instead a own was intended to encourage to not use this class in all cases.
There is a reason why this simple class is not included in the java.lang or even guava
https://github.com/google/guava/wiki/IdeaGraveyard
https://mail.openjdk.java.net/pipermail/core-libs-dev/2010-March/003995.html

But I'm also ok with it to have this in a separatet java source file.

tyrasd: actually, after a second look, I realized that this class (and previously, the `MutablePair<>`s used in this class) is only used to store pairs of values (generic type `X`) that are associated with an index (generic type `U`) in an internal `MapReducer` object (see above in https://github.com/GIScience/oshdb/pull/79/files#diff-5c05e36796aa82dfa133e2fa41932bb2R87). So, this class is totally fine to live here and can even be made private

PR #82, opened 2019-03-01
explicitly contruct cellid ranges iterable

tyrasd:
instead of automatically through a lambda to avoid catching too much stuff (e.g. parent classes), which can cause serialization issues

### DIFF
--- a/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGridTree.java
+++ b/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGridTree.java
@@ -42,6 +42,7 @@ public class XYGridTree implements Serializable {
    * @param latitude
    * @return An iterator over the cellIds in all zoomlevel
    */
+  @SuppressWarnings("Convert2Lambda")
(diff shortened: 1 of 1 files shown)


oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGridTree.java
```diff
@@ -205,35 +207,42 @@ public boolean equals(Object obj) {
...
+   * @param bbox
    * @param enlarge
    * @return
    */
-  public Iterable<CellIdRange> bbox2CellIdRanges(final OSHDBBoundingBox BBOX, final boolean enlarge) {
-    return (Iterable<CellIdRange>) () -> new Iterator<CellIdRange>() {
-      private int level = 0;
-      private Iterator<IdRange> rows = gridMap.get(level).bbox2CellIdRanges(BBOX, enlarge).iterator();
-
+  @SuppressWarnings("Convert2Lambda")
```
rtroilo: Is this intelij specific?

tyrasd: Probably.

PR #111, opened 2019-03-07
speedup building of geometries from osm entities

tyrasd:
includes two improvements for building geometries:

* **speedup building of complex multipolygons**
  Relations with many (small) inner rings used to take a long time to assemble, because each of the expensive `within` operation in a loop. By using a prepared geometry, this is substantially faster. Eespecially for particularly complex example like [relation/571669](https://www.openstreetmap.org/relation/571669) this makes a large difference: processing times went down from several minutes to a few seconds.
* **avoid clipping if not necessary**
  This applies the idea from #66 to the cellIterator functions: Some objects might have a large OSH bbox, but small (snapshot or contribution) geometries. Then it is not always necessary to calculate actual intersections, which are potentially slow computations.


tyrasd:
selection_214

SlowMo24:
can confirm speed improvement of 3x faster on ignite (tested locally)

PR #131, opened 2019-03-13
Fix: rename GridOSHEntity to GridOSHEntities

SlowMo24:
Merge this branch into master **after** serialization has been changed from serializing classes to serializing data.

Being a collection of entities GridOSHEntities should be plural
according to our coding style.

See https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb/merge_requests/17


SlowMo24:
related? #135

SlowMo24:
Fails because https://github.com/GIScience/oshdb/blob/rename-GridOSHEntity/oshdb-util/src/test/resources/test-data.mv.db holds GridOSHEntiy not GridOSHEntities
needs to be updated before merging!

rtroilo:
see new issue #137

SlowMo24:
So the class GridOSHEntities would not exist any more?

joker234:
outdated, closed as suggested before

PR #158, opened 2019-04-12
check timestamps when aggregating

tyrasd:
If the `aggregateByTimestamps(callback)` function returns timestamps outside of the query's time range, an explicit exception will now be thrown.

Before this case was not checked and did in some cases to hard to debug `NullPointerException`s in various places (e.g. in [`OSHDBCombinedIndex.compareTo`](https://github.com/GIScience/oshdb/blob/master/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/generic/OSHDBCombinedIndex.java#L31-L34), or potentially even inconsistent results.


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -173,9 +174,18 @@ private MapAggregator(MapAggregator<U, ?> obj, MapReducer<IndexValuePair<U, X>>
...
   ) {
     final TreeSet<OSHDBTimestamp> timestamps = new TreeSet<>(this.mapReducer.tstamps.get());
+    final OSHDBTimestamp minTime = timestamps.first();
+    final OSHDBTimestamp maxTime = timestamps.last();
     return this.aggregateBy(data -> {
       // match timestamps to the given timestamp list
-      return timestamps.floor(indexer.apply(data));
+      OSHDBTimestamp aggregationTimestamp = indexer.apply(data);
+      if (aggregationTimestamp.compareTo(minTime) < 0
+          || aggregationTimestamp.compareTo(maxTime) > 0) {
```
rtroilo, line 183: `indexer.apply(data)` could potential return a null value.  `.aggregateByTimestamp(x -> null) `.
so maybe also at a check for that case to
```java
if ( aggregationTimestamp == null 
     || aggregationTimestamp.compareTo(minTime) < 0
     || aggregationTimestamp.compareTo(maxTime) > 0) {
```

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -837,9 +838,18 @@ protected MapReducer(MapReducer<?> obj) {
...
   ) throws UnsupportedOperationException {
     final TreeSet<OSHDBTimestamp> timestamps = new TreeSet<>(this.tstamps.get());
+    final OSHDBTimestamp minTime = timestamps.first();
+    final OSHDBTimestamp maxTime = timestamps.last();
     return new MapAggregator<OSHDBTimestamp, X>(this, data -> {
       // match timestamps to the given timestamp list
-      return timestamps.floor(indexer.apply(data));
+      OSHDBTimestamp aggregationTimestamp = indexer.apply(data);
+      if (aggregationTimestamp.compareTo(minTime) < 0
+          || aggregationTimestamp.compareTo(maxTime) > 0) {
```
rtroilo, line 847: same case as for the MapAggregator class.
`indexer.apply(data)` could return a null value, so maybe add this check too

PR #174, opened 2019-05-22
update documentation to 0.5.1 release

tyrasd:
no changes in content, since it's mostly a bugfix release.


SlowMo24:
hm, this seems improvable. Would links to 'latest' and 'stable' improve the process?

tyrasd:
> Would links to 'latest' and 'stable' improve the process?

yeah, that would indeed make this process a bit easier. On the other hand, these documents/link have to be checked for major releases anyway, since the links could become broken (e.g. when classes/packages are renamed, or methods deprecated).

PR #194, opened 2019-07-17
Updated mail addresses for authors

joker234:
Something to think about, before merging:
 * remove authors completely from source code, instead of this PR, and use only the authors in the [parent](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/parent/blob/master/pom.xml)
 * if the [parent](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/parent/blob/master/pom.xml) gets updated, we should release it first and change the parent version here, too

### DIFF
--- a/oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/util/OSHDB2Ignite.java
+++ b/oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/util/OSHDB2Ignite.java
@@ -29,9 +29,6 @@ import org.heigit.bigspatialdata.oshdb.util.TableNames;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- * @author Rafael Troilo &lt;rafael.troilo@heigit.org&gt;
- */
 public class OSHDB2Ignite {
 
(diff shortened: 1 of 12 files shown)


tyrasd:
IMHO we don't need author information in the javadoc/sources. We have a SCM to actually track code authorship. We should urge people to update their git mail addresses, though.

A similar argument could also be applied to the authors information in the parent pom, but I'm not 100% sure about that case.

joker234:
> IMHO we don't need author information in the javadoc/sources. We have a SCM to actually track code authorship. We should urge people to update their git mail addresses, though.

Totally agree. So we should remove them.

> A similar argument could also be applied to the authors information in the parent pom, but I'm not 100% sure about that case.

I think we need them (or at least one person) for the releases (see [OSSRH central requirements](https://central.sonatype.org/pages/requirements.html#developer-information) for details). I don't have any problems to store and update these information on [one single place](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/parent/blob/master/pom.xml#L35) for our group. Maybe we should distinguish more between developers and contributors, see [Maven Model Definiton](http://maven.apache.org/ref/3.3.3/maven-model/maven.html#developer).

joker234:
Tidied up authors as suggested and bumped parent version. Can be merged now.

PR #198, opened 2019-09-24
Removing UNKNOWN from OSMType, throw Exception instead!

rtroilo:
## Type of change
- [x] New feature (non-breaking change which adds functionality)

## Description
- Remove type UNKOWN from OSMType and makes more use of the default case in switch statements instead

### DIFF
--- a/oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/extract/TypeStartFinder.java
+++ b/oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/extract/TypeStartFinder.java
@@ -196,7 +196,7 @@ public class TypeStartFinder {
 			return OSMType.WAY;
 		if (group.getRelationsCount() > 0)
 			return OSMType.RELATION;
-		return OSMType.UNKNOWN;
+		return null;
 
 	}
 
--- a/oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/extract/TypeStartFinder.java
+++ b/oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/extract/TypeStartFinder.java
@@ -20,184 +20,183 @@ import crosby.binary.Osmformat;
 
 public class TypeStartFinder {
 
-	public static OsmPbfMeta getMetaData(Path pbf) throws InvalidProtocolBufferException {
-		OsmPbfMeta meta = new OsmPbfMeta();
-		meta.pbf = pbf;
-		Path metaPath = pbf.getParent().resolve(pbf.getFileName().toString() + ".meta");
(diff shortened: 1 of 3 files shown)


oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/xmlreader/OSMXmlReader.java
```diff
@@ -206,7 +206,7 @@ else if ("way".equalsIgnoreCase(type)) {
           } else if ("relation".equalsIgnoreCase(type)) {
             t = OSMType.RELATION;
           } else {
-            t = OSMType.UNKNOWN;
+        	t = null;
```
tyrasd: tab -> 2 spaces please ;)

oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/extract/TypeStartFinder.java
```diff
@@ -196,7 +196,7 @@ public static OSMType getType(PbfBlob blob) throws InvalidProtocolBufferExceptio
 			return OSMType.WAY;
 		if (group.getRelationsCount() > 0)
 			return OSMType.RELATION;
-		return OSMType.UNKNOWN;
+		return null;
```
tyrasd: why do you not throw an exception like in the `fromInt` method below? Can't a `null` cause hard to understand `NullPointerException`s down the line?

tyrasd:
ping?

I agree that this would be a good change for code clarity. For example, it would have allowed to write https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter/blob/51227047/src/main/java/org/heigit/ohsome/filter/TypeFilter.java#L41-42 in a more simplified way. :wink: 

But there are still two minor comments that need to be clarified before we can go on with merging this. :smiley:

rtroilo:
I'm not sure which minor comments you refer, but I think I solved them already
> why do you not throw an exception
Now there is an IllegalArgumentException instead of a NullPointerException

And thank you for fixing the Typo ;-)

oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/extract/TypeStartFinder.java
tyrasd: here would be a place where a switch statement dereferences a possibly `null` OSMType value: https://github.com/GIScience/oshdb/blob/remove-unknown-type/oshdb-tool/etl/src/main/java/org/heigit/bigspatialdata/oshdb/tool/importer/extract/TypeStartFinder.java#L52

rtroilo: should be fixed now, it throws an exception instead of returning null, if getType could not find out the current OSMType.

PR #202, opened 2019-11-14
Improve data streaming, part 2

tyrasd:
follow up for #200:

Makes queries over a large area (e.g. global ones) significantly faster.

Also includes an experimental implementation for streaming using the scanquery backend. It can work well for large/global queries of sparsely distributed features.

# Checklist

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteAffinityCall.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteAffinityCall.java
@@ -1,11 +1,18 @@
 package org.heigit.bigspatialdata.oshdb.api.mapreducer.backend;
 
 import com.google.common.collect.Streams;
+import com.google.common.primitives.Ints;
 import java.io.IOException;
 import java.sql.SQLException;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.TreeMap;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.LongStream;
@@ -15,6 +22,9 @@ import org.apache.ignite.Ignite;
 import org.apache.ignite.IgniteCache;
 import org.apache.ignite.IgniteCompute;
 import org.apache.ignite.IgniteException;
+import org.apache.ignite.binary.BinaryObject;
+import org.apache.ignite.cache.query.QueryCursor;
+import org.apache.ignite.cache.query.ScanQuery;
 import org.apache.ignite.lang.IgniteCallable;
 import org.apache.ignite.lang.IgniteFuture;
 import org.apache.ignite.lang.IgniteFutureTimeoutException;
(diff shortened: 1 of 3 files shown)


tyrasd:
some more relevant details to this PR:

* the "pure" ScanQuery [implementation](https://github.com/GIScience/oshdb/pull/202/files#diff-4263d0dc6ced930075faf0738987dc88R568-R598) is quite limited, because it 1) doesn't support timeouts and 2) doesn't handle backpressure of data: data is sent immediately and in full to the ignite client as it is processed on the workers, where it is buffered. When the client cannot process the incoming data fast enough, the buffer blows up the stack on the client. This means that the scanquery backend only works if the resulting data is expected to be small, or if one is sure that the client can handle the incoming data faster than the workers can deliver it (which is typically not the case for simple data extraction requests).
* the implementation from #200 was slow for global requests, for the same reason that the _LocalPeek_ backend is slow for global request: all potential data cells need to be tested for their existence. This PR implements an alternative preflight request, which is used for large requests. See https://github.com/GIScience/oshdb/pull/202/files#diff-3c45747da461877b2884fa4e19705e96R236-R239 for an explaination of the assumptions this heuristic is based on.
* the result of the preflight requests (i.e. a `Set` of cell ids) could also be more compactly stored in a RoaringBitMap (which we already use in other parts of the oshdb), but this failed for be because serialization issues. Normally, this shouldn't be the bottleneck of a typical request, though.

richardstartin:
@tyrasd 

> the result of the preflight requests (i.e. a Set of cell ids) could also be more compactly stored in a RoaringBitMap (which we already use in other parts of the oshdb), but this failed for be because serialization issues.

Can you share any details about this?

tyrasd:
@richardstartin it's quite odd, here's two ("minimal") examples:

Regular `RoaringBitmap` – here, a broadcast job is sent to a remote Ignite cluster, which returns a RoaringBitmap object as a result. The result is expected to be serialized on the cluster side and be deserialized in its "original" state on the client, but isn't:

```java
    RoaringBitmap r = new RoaringBitmap();
    r.add(0,1,2,3);

    System.out.println(r.toString());
    // prints {0,1,2,3} = correct

    Ignition.setClientMode(true);
    Ignite ignite = Ignition.start("/…/ignite-config.xml");
    ignite.compute().broadcast(() -> {
      RoaringBitmap r2 = new RoaringBitmap();
      r2.add(0,1,2,3);
      return r2;
    }).stream().forEach(System.out::println);
    // prints {0,16,0,1} = incorrect

    ignite.close();
```

When using the 64-bit extension (`Roaring64NavigableMap`), it gets even worse:

```java
    Roaring64NavigableMap r = new Roaring64NavigableMap();
    r.add(0,1,2,3, 1L << 41, 1L << 43);

    System.out.println(r.toString());
    // prints {0,1,2,3,2199023255552,8796093022208} = correct

    Ignition.setClientMode(true);
    Ignite ignite = Ignition.start("/…/ignite-config.xml");
    ignite.compute().broadcast(() -> {
      Roaring64NavigableMap r2 = new Roaring64NavigableMap();
      r2.add(0,1,2,3, 1L << 41, 1L << 43);
      return r2;
    }).stream().forEach(System.out::println);
    // throws an exception -> see below

    ignite.close();
```

this results in the following exception being thrown:

```
Exception in thread "main" class org.apache.ignite.binary.BinaryObjectException: Failed to unmarshal object with optimized marshaller
	at org.apache.ignite.internal.binary.BinaryUtils.doReadOptimized(BinaryUtils.java:1765)
	…
Caused by: class org.apache.ignite.IgniteCheckedException: Failed to deserialize object with given class loader: [clsLdr=jdk.internal.loader.ClassLoaders$AppClassLoader@3d4eac69, err=Failed to deserialize object [typeName=org.roaringbitmap.longlong.Roaring64NavigableMap]]
	at org.apache.ignite.internal.marshaller.optimized.OptimizedMarshaller.unmarshal0(OptimizedMarshaller.java:237)
	…
Caused by: java.io.IOException: Failed to deserialize object [typeName=org.roaringbitmap.longlong.Roaring64NavigableMap]
	at org.apache.ignite.internal.marshaller.optimized.OptimizedObjectInputStream.readObject0(OptimizedObjectInputStream.java:350)
	…
Caused by: java.io.IOException: org.roaringbitmap.InvalidRoaringFormat: I failed to find a valid cookie.
	at org.roaringbitmap.InvalidRoaringFormat.toIOException(InvalidRoaringFormat.java:36)
	at org.roaringbitmap.RoaringBitmap.deserialize(RoaringBitmap.java:1735)
	at org.roaringbitmap.longlong.Roaring64NavigableMap.deserialize(Roaring64NavigableMap.java:1146)
	at org.roaringbitmap.longlong.Roaring64NavigableMap.readExternal(Roaring64NavigableMap.java:906)
	at org.apache.ignite.internal.marshaller.optimized.OptimizedObjectInputStream.readExternalizable(OptimizedObjectInputStream.java:555)
```
 
That's basically where I gave up, since in this specific use-case the mentioned `Set<Long>` objects don't get particularly large to cause issues. :man_shrugging: 
Should this be reported to the https://github.com/RoaringBitmap/RoaringBitmap/issues or is it a known issue?

richardstartin:
@tyrasd thanks for the info. I don't have the full background as to why standard serialisation mechanisms aren't used but there are two APIs, the most efficient de/serialises directly from/to a `ByteBuffer`, and a slower API accepting `DataInput` and `DataOutput`. I don't think it should just blow up like this though.

tyrasd:
thanks, I'll keep it in mind.

PR #204, opened 2019-12-17
fix (literal) edge cases in fast-polygon operations class

tyrasd:
Literal edge case here :laughing: – Sometimes during the preprocessing step (responsible for optimizing the polygon operations) the intermediate results may contain superfluous points or lines along the clipping edge(s). These additional geometries cause problems further down the processing pipeline. Since these edge artifacts don't contribute anything to the final oshdb query result, they can just be filtered out.

Here is an example of a geometry which can cause such a situation: https://gist.github.com/tyrasd/19ec9399d42dcf76ed7ab16432052afe (found by @redfrexx ~ thanks!).
 
# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,13 @@ Changelog
 
 ## 0.6.0 SNAPSHOT (current master)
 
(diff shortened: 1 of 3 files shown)


tyrasd:
> the gcPolys list could hold also only a single polygon?

it can. good point. I'll add a path for the single polygon case.

PR #207, opened 2020-01-16
Update version of first-steps documentation

SlowMo24:
We should look for other places where this is needed and possibly create a list to not forget at the next update.


tyrasd:
there are a couple more places, especially in links to our javadoc: https://github.com/GIScience/oshdb/commit/9f6e2f0e17b16f11885633f82f07e2eb92b8af58

PR #215, opened 2020-02-26
give timeout tests a better chance to actually fail

tyrasd:
Sometimes test runners are so quick that they sometimes do not reach the specified timeout during (some of) the queries in the timeout tests.

Example of a failed test run because of this: https://jenkins.ohsome.org/blue/organizations/jenkins/oshdb/detail/master/266/pipeline#step-92-log-477

This PR should prevent this by giving the queries a more realistic chance to actually time out by:

* setting the timeout to 0ms (from 1ms)
* letting the OSHDB query work through slightly more data (from just 1 entity filtered by id, to all in the test data set)


tyrasd:
damn, it seems like this still didn't fix the issue: https://jenkins.ohsome.org/job/oshdb/job/master/271/ :unamused:

PR #217, opened 2020-02-27
 give timeout tests a better chance to actually fail, part II

tyrasd:
addendum to #215 (see https://github.com/GIScience/oshdb/pull/215#issuecomment-591514181) – thanks @rtroilo for the suggested fix :smiley:


tyrasd:
> looks good, but could we extract this in a general delay method?

seems like a good idea. Any preferences where this method should live? It's probably mostly (only?) useful for tests, but maybe eventually not just for the map-reducer tests where it would be used at the moment. :thinking: 

@rtroilo would you like to add this method (just push into the `fix-timeout-tests` branch)?

PR #220, opened 2020-04-01
prepare release 0.5.6

tyrasd:


tyrasd:
btw: here's a useful command for upgrading the links to the javadoc in the documentation:

```bash
sed -i 's!docs.ohsome.org/java/oshdb/0.5.5!docs.ohsome.org/java/oshdb/0.5.6!g' `grep "0.5.5" -rl`
```

PR #230, opened 2020-04-29
avoid using IntelliJ/IDEA style suppress warnings comment

tyrasd:
This is to improve interoperability between different IDEs. For example Eclipse, does not support the IDEA/IntelliJ syntax `//noinspection unchecked`. If possible, unchecked casts should be avoided anyway. But where it cannot be avoided, one can (probably almost always) rewrite unchecked casts to use the regular Java annotation syntax `@SuppressWarnings("unchecked")`.

For example, instead of

```
//noinspection unchecked …bla…bla…
return (UncheckedType) …;
```

one can write it as:

```
@SuppressWarnings("unchecked") // …bla…bla…
UncheckedType result = (UncheckedType) …;
return result;
```

### todo
- [x] adapt current code: drop and rewrite [all 37 current occurrences](https://jenkins.ohsome.org/job/oshdb/view/change-requests/job/PR-230/1/checkstyle/category.-1850931927/) of `//noinspection unchecked`


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -759,7 +762,8 @@ public X setValue(X value) {
   @Contract(pure = true)
   public <R> MapAggregator<U, R> map(SerializableFunction<X, R> mapper) {
     return this.copyTransform(this.mapReducer.map(inData -> {
-      //noinspection unchecked – trick/hack to replace mapped values without copying pair objects
+      @SuppressWarnings("unchecked")
+      // trick/hack to replace mapped values without copying pair objects
```
rtroilo, line 766: Usually I prefer to have the @SuppressWarnings directly following the affected line. 
So in this sense the comment would come before or in the same line as  @SuppressWarings:
```
// trick/hack to replace mapped values without copying pair objects
@SuppressWarnings("unchecked")
IndexValuePair<U,R> outData = (IndexValuePair<U,R>)inData;
```
But this is only my personal preference and does not hinder the approval!

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -1836,8 +1851,8 @@ protected TagTranslator getTagTranslator() {
 
   // hack, so that we can use a variable that is of both Geometry and implements Polygonal (i.e.
   // Polygon or MultiPolygon) as required in further processing steps
+  @SuppressWarnings("unchecked") // all setters only accept Polygonal geometries
```
rtroilo, line 1854: this is how I prefer it, comments above or in the some line. (see previous comment)

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
tyrasd: :thinking:  hmm, I did it like this because it is otherwise not clear that the comment is related to the `@SuppressWarnings` statement: When the comment is in the line above, it could also be just a regular inline comment. Short comments about the suppress statement you would also put _after_ the annotation (`@SuppressWarnings("unchecked") // bla bla`), wouldn't you?

But I guess that's ultimately a matter of taste. :shrug:

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
tyrasd: I also prefer this version, but sometimes the line becomes too long for the space left after the annotation, then I just chose to put it in a new line, so the comment is without a line break (thus, slightly easier to read): I prefer `@Suppress… \n // long bla` over `@Suppress… // long \n // bla`. But again, this might just be a matter of personal taste.

PR #238, opened 2020-05-12
check if VERSION could be correctly determined

tyrasd:
Under some error conditions `mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate` can fail. This is for example the case when a module's parent module cannot be loaded from the specified repositories. Before this fix, this lead to the variable `VERSION` to be empty, resulting in any (snapshot) release deploys to be skipped silently. This changes the behaviour to instead fail early in the _build and test_ step, so one can see that the build didn't go as planned.

//edit: I also removed some unused variables. these can be added again later if they are going to be needed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,7 +38,6 @@ pipeline {
           }
         }
         script {
-
           server = Artifactory.server 'HeiGIT Repo'
           rtMaven = Artifactory.newMavenBuild()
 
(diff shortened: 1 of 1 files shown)


tyrasd:
Thanks @SlowMo24! I like your version much better, because it also triggers when the version can be read from the pom, but would contain typos, unexpected characters and/or non semantic version numbers. :+1:

Issue #225, opened 2020-04-07
get rid of global keytables


tyrasd:
Our current global keytables approach has many downsides: they are bulky, slow, unintuitive and cause headaches when implementing code that needs to be executed on a distributed system (because contrary to the rest of the data, the keytables are actually _not_ distributed).

Maybe a better way to solve the compression of OSM tags would be to get rid of the whole concept of separate global keytables altogether. I would prefer to try a solution with small local string tables, as it is solved in other state of the art OSM data schemas. For example like the `StringTable` in an [osm-pbf data block](https://wiki.openstreetmap.org/wiki/PBF_Format#Definition_of_OSMData_fileblock), or like the attributes in [MVT](https://docs.mapbox.com/vector-tiles/specification/#encoding-attributes). I believe [Apache ORC](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+ORC#LanguageManualORC-StringColumnSerialization) does something similar with strings.

Hagellach37:
not planned for now.

PR #247, opened 2020-05-29
⚠️  close tagtranslators in mapreducers

tyrasd:
⚠️ – 292c011 has breaking changes, needs to be discussed thoroughly
~🚧 – merge after #246~

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java
@@ -181,6 +181,17 @@ public abstract class MapReducer<X> implements
   @NotNull
   protected abstract MapReducer<X> copy();
 
+  private void close() {
+    if (this.tagTranslator != null) {
+      try {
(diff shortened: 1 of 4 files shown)


rtroilo:
Moving the ownership of the TagTranslator from MapReducer to OSHDBDatabase could solve the problem.
Up to now I don't see a case where we need to have the same OSHDBDatabase connection with different TagTranslator instances.

```java
// set explicit  keytables connection for Cluster OSHDBDatabases
try(Connection keytables = DriverManager.getConnection(url);
    OSHDBDatabase oshdb = new OSHDBIgnite("ignite.xml", keytables);){
  // ...
}

try(Connection conn = DriverManager.getConnection(url);
    OSHDBDatabase oshdb = new OSHDBJdbc(conn)){
  //  ...
}
```
The MapReducer instance gets the TagTranslator reference through the OSHDBDatabase and the OSHDBDatabase is responsible for closing the TagTranslator probably.

This also has the positive side effect that the TagTranslator could be better utilize its cache tags by executing multiple MapReducer with the some OSHDBDatabase.

tyrasd:
related: #225 :pray:

Issue #129, opened 2019-03-11
OSHDBGeometryBuilder - touching inner rings


kathap:
* [x] [Test750](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/750):
The result is not valid OGC polygon
![image10](https://user-images.githubusercontent.com/30441792/54133065-b00aa380-4415-11e9-9c69-a4a5d0ff196b.png)
* [x] [test751](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/751):
The result is not valid OGC polygon
![image11](https://user-images.githubusercontent.com/30441792/54133080-b8fb7500-4415-11e9-87eb-71db599cea3f.png)

* [x] [test783](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/783):
The result is not valid OGC polygon
![image12](https://user-images.githubusercontent.com/30441792/54133127-cf093580-4415-11e9-8a8c-18dcb564b4cd.png)

* [x] [test784](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/784):
The result is not valid OGC polygon
![image13](https://user-images.githubusercontent.com/30441792/54133215-f3fda880-4415-11e9-8fbf-df269f53321b.png)

* [x] [test785](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/785):
The result is not instance of MultiPolygon
![image14](https://user-images.githubusercontent.com/30441792/54133238-ff50d400-4415-11e9-9521-8412de0f868b.png)

[code for these tests](https://github.com/GIScience/oshdb/blob/osm-history-testdata/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java)

Issue #123, opened 2019-03-11
OSHDBGeometryBuilder - Multipolygon with two outer rings and two inner rings touching in two nodes


kathap:
* [x] [Test 777](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/777)
* ~~[Test 778](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/778)~~
* [x] [Test 779](https://github.com/osmcode/osm-testdata/tree/master/grid/data/7/779)

![image1](https://user-images.githubusercontent.com/30441792/54130511-ccf0a800-4410-11e9-8982-5d7aab3e7efb.png)

[code for these tests](https://github.com/GIScience/oshdb/blob/osm-history-testdata/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java)

tyrasd:
I believe 778 is not really a valid test case.

According to https://wiki.openstreetmap.org/wiki/Relation:multipolygon: “[inner ways make] up the optional inner ring(s) delimiting the excluded holes that must be fully inside the area delimited by outer ring(s).”

For overlapping inner rings (such as it is the case here), it is not defined if the geometry building algorithm should use the even-odd rule, the non-zero-winding rule, a subtractive algorithm or something else.

PR #249, opened 2020-06-17
dedicated handling of touching inner rings in multipoygons

tyrasd:
The previous workaround of using `buffer(0)` to let JTS join any touching inner rings can be very slow, up to the point of potentially bringing a query to an almost complete halt when it has to be called on very large multipolygons (e.g. `relation/1205151`).

This adds dedicated handling of touching inner rings, and merges them before the geometry object is created.

A few test cases were adjusted, because the new handling is following more strictly the multipolygon validity rules given in the osm wiki. (related: #123, #129)

Also includes some minor code cleanup (e.g. adding of a private construtor to the static geometry builder utility class to hide the implicit public constructor).

# Checklist

### DIFF
--- a/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
+++ b/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
@@ -460,7 +460,13 @@ public class OSHDBGeometryBuilder {
     Geometry geom = OSHDBGeometryBuilder.getGeometry(entity, timestamp, areaDecider);
     return Geo.clip(geom, clipPoly);
   }
-  
+
+  /**
+   * Converts a JTS bounding box ("envelope") to an OSHDBBoundingBox object.
+   *
+   * @param envelope the bounding box object to convert
+   * @return the same bounding box as an OSHDBBoundingBox object
+   */
   public static OSHDBBoundingBox boundingBoxOf(Envelope envelope) {
     return new OSHDBBoundingBox(
         envelope.getMinX(),
--- a/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
+++ b/oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
@@ -3,8 +3,10 @@ package org.heigit.bigspatialdata.oshdb.util.geometry;
 import com.google.common.collect.Lists;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -368,47 +449,84 @@ private static Polygon constructMultipolygonPart(
...
+        return otherSegment.id1 == this.id1 && otherSegment.id2 == this.id2
+            || otherSegment.id1 == this.id2 && otherSegment.id2 == this.id1;
+      } else {
+        return super.equals(other);
+      }
+    }
+
+    @Override
+    public int hashCode() {
+      return ((int) this.id1) + ((int) this.id2);
```
joker234, line 529: Are the here used ids always hashes? Is the addition unique enough?

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java
```diff
@@ -1109,17 +1113,20 @@ public void test784() throws ParseException {
...
     // Valid OSM multipolygon with two touching inner rings leaving an empty area.
     OSMEntity entity = testData.relations().get(785900L).get(0);
     Geometry result = OSHDBGeometryBuilder.getGeometry(entity, timestamp, tagInterpreter);
-    assertTrue(result instanceof MultiPolygon);
-    assertTrue(result.isValid());
-    assertEquals(2, result.getNumGeometries());
-    assertEquals(1,
-        ((Polygon)result.getGeometryN(0)).getNumInteriorRing()
-        + ((Polygon)result.getGeometryN(1)).getNumInteriorRing()
-    );
```
joker234, line 1122: Why is the check for multiple interior rings not necessary anymore?

joker234 (commented):
Some open points need to be considered, mostly questions or suggestions.

joker234:
@tyrasd I checked all finished points of your checklist above. Can you think about whether we need to do the remaining ones (benchmarks, changelog, examples, documentation)

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -285,14 +285,95 @@ private static Geometry getMultiPolygonGeometry(
...
+          || ring.getFirst().getId() == segment.id2 && ring.getLast().getId() == segment.id1) {
+        return;
+      }
+      ring.add(ring.removeFirst());
+    }
+  }
+
+  private static void mergeSegmentsToRing(LinkedList<OSMNode> target, LinkedList<OSMNode> source) {
+    if (target.getFirst().getId() == source.getFirst().getId()) {
+      Collections.reverse(source);
```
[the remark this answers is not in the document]
tyrasd: you mean having four `while` loops (two here and the current two in the `else` branch) to check all possible joining possibilities? that would be more performant, I guess :+1:

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
tyrasd: The ids are just the osm ids of the nodes of the respective way segment. I think I just wanted something quick satisfies equality regardless of the order of the ids. Could this result in issues down the line, e.g. when used in a HashMap?

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java
```diff
@@ -962,15 +961,21 @@ public void test776() throws ParseException {
...
+    should also be considered an invalid multipolygon relation.”
+
+    Here, the inners form a ring around a portion of the interior.
+    */
+    // https://github.com/GIScience/oshdb/issues/123
     // Multipolygon with two outer rings and two inner rings touching in two nodes
     OSMEntity entity = testData.relations().get(777900L).get(0);
     Geometry result = OSHDBGeometryBuilder.getGeometry(entity, timestamp, tagInterpreter);
-    assertTrue(result instanceof MultiPolygon);
-    assertTrue(result.isValid());
```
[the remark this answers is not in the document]
tyrasd: since the test777 is according to the explanation above not a valid test case, because the touching inner rings are not `valid` in the OGC definition and that's exactly what `isValid` tests. The other asserts below don't make sense because they relied on some "magic" fixing the invalidity of the input data. :shrug:

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java
tyrasd: similar argument as above: we don't try to magically fix the invalid input data anymore, so there's no guarantee in the outcome of such input data.

tyrasd:
> @tyrasd I checked all finished points of your checklist above. Can you think about whether we need to do the remaining ones (benchmarks, changelog, examples, documentation)

I would definitely add this to teh changelog. benchmarks would be nice, but let's see if we find time. documentation/examples are probably not needed, since the API stayed the same.

oshdb-util/src/main/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilder.java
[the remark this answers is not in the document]
tyrasd: ok, I took a second look. "some" reversing the list is necessary anyway, because in order for the two lines `target`/`source` to be joined into a proper ring at the end of this method, they need to have the correct order anyway. One might save a few swap operations by doing this at the end of the method, but I believe it's at this point an unnecessary optimization and not worth the added complexity in the code.

joker234 (approved):

Issue #241, opened 2020-05-15
rename packages to match updated groupId


joker234:
Open TODOs from #234:
See [comment](https://github.com/GIScience/oshdb/pull/234#pullrequestreview-407701110).

- Move sources `src/main/java/org/heigit/bigspatialdata` → `src/main/java/org/heigit/ohsome`
- Move tests `src/test/java/org/heigit/bigspatialdata` → `src/test/java/org/heigit/ohsome`
- Rename imports
- Better use `grep` and `find` after these changes again, if I missed anything.

joker234:
Don't forget to fix [oshdb-examples#1](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/issues/1) when this is fixed.

tyrasd:
In principle all dependent repositories need to be updated (ohsome-api, ohsome-filter, oshdb-benchmarks, etc.)

tyrasd:
postponed to 0.7, because 

1. it would have made any "0.5" H2 files incompatible (which work otherwise still fine with 0.6)
2. upgrading the ohsome-filter library would have been difficult. -> will be solved by integrating into this repository (see https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter/-/issues/15)

PR #250, opened 2020-06-18
Improve documentation of group by entity feature

tyrasd:
additionally, this makes `OSMContribution` and `OSMEntitySnapshot` objects comparable to each other, defining a natural order: first by their respective entity _type_ and _id_, then by their _timestamp_.

# Checklist
- [ ] My code follows the [code-style](https://github.com/GIScience/oshdb/blob/master/CONTRIBUTING.md) rules and I have checked on the [static analyses](https://jenkins.ohsome.org/job/oshdb/view/change-requests/) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) (if applicable) results
- [ ] I have commented my code
- [ ] I have written javadoc (required for public methods)
- [ ] I have added sufficient unit tests
- [ ] I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/master/documentation)
- [ ] I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) if necessary

### DIFF
--- a/documentation/manual/views.md
+++ b/documentation/manual/views.md
@@ -54,7 +54,7 @@ Note that there exist [cases](https://github.com/GIScience/oshdb/issues/87) wher
 GroupByEntity
 -------------
 
-The [`groupByEntity()`](https://docs.ohsome.org/java/oshdb/0.5.10/aggregated/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.html#groupByEntity--) method of a MapReducer slightly changes the way the MapReducers recieves and transforms values: Instead of iterating over each snapshot or contribution individually, in this mode all snapshots or all contributinos of an individual OSM entity are collected into a list of values first. This makes it possible to investigate the full edit history of individual OSM objects at once.
+The [`groupByEntity()`](https://docs.ohsome.org/java/oshdb/0.5.10/aggregated/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.html#groupByEntity--) method of a MapReducer slightly changes the way the MapReducers recieves and transforms values: Instead of iterating over each snapshot or contribution individually, in this mode all snapshots or all contributions of an individual OSM entity are collected into a (by timestamp sorted) list first. This makes it possible to investigate the full edit history of individual OSM objects at once, which is for example needed when one is looking for contributions that got reverted at a later point in time.
 
 It is recommended to call this method immediately after creating the MapReducer from a view:
 
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/object/OSMContribution.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/object/OSMContribution.java
@@ -267,6 +267,11 @@ public class OSMContribution implements OSHDBMapReducible, Comparable<OSMContrib
     return data.changeset;
   }
 
+  /**
+   * {@inheritDoc}
+   *
+   * Note: this class has a natural ordering that is inconsistent with equals.
+   */
   @Override
   public int compareTo(@Nonnull OSMContribution other) {
     return ComparisonChain.start()
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/object/OSMContribution.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/object/OSMContribution.java
@@ -1,7 +1,9 @@
 package org.heigit.bigspatialdata.oshdb.api.object;
 
+import com.google.common.collect.ComparisonChain;
 import java.util.EnumSet;
 import java.util.Objects;
+import javax.annotation.Nonnull;
 import org.heigit.bigspatialdata.oshdb.osh.OSHEntity;
 import org.heigit.bigspatialdata.oshdb.osm.OSMEntity;
 import org.heigit.bigspatialdata.oshdb.osm.OSMRelation;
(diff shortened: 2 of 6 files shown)


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/object/OSMContribution.java
```diff
@@ -264,4 +266,25 @@ public int getContributorUserId() {
...
+  @Override
+  public boolean equals(Object other) {
+    if (!(other instanceof OSMContribution)) {
+      return false;
+    }
+    OSMContribution otherContribution = (OSMContribution) other;
+    return this.getOSHEntity().equals(otherContribution.getOSHEntity())
+        && Objects.equals(this.getEntityBefore(), otherContribution.getEntityAfter())
+        && Objects.equals(this.getEntityAfter(), otherContribution.getEntityBefore())
+        && this.getTimestamp().equals(otherContribution.getTimestamp());
```
SlowMo24, line 288: would checking the timestamp first be faster?

SlowMo24: also with snapshot?

documentation/manual/views.md
```diff
@@ -54,12 +54,12 @@ Note that there exist [cases](https://github.com/GIScience/oshdb/issues/87) wher
...
 
-The [`groupByEntity()`](https://docs.ohsome.org/java/oshdb/0.5.9/aggregated/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.html#groupByEntity--) method of a MapReducer slightly changes the way the MapReducers recieves and transforms values: Instead of iterating over each snapshot or contribution individually, in this mode all snapshots or all contributinos of an individual OSM entity are collected into a list of values first. This makes it possible to investigate the full edit history of individual OSM objects at once.
+The [`groupByEntity()`](https://docs.ohsome.org/java/oshdb/0.5.9/aggregated/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.html#groupByEntity--) method of a MapReducer slightly changes the way the MapReducers recieves and transforms values: Instead of iterating over each snapshot or contribution individually, in this mode all snapshots or all contributions of an individual OSM entity are collected into a (by timestamp sorted) list first. This makes it possible to investigate the full edit history of individual OSM objects at once, which is for example needed when one is looking for contributions that got reverted at a later point in time.
```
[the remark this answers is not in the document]
tyrasd: the links are updated semi-automatically (see step 4 in https://confluence.gistools.geog.uni-heidelberg.de/display/oshdb/how+to+create+a+release): since the creation of a release requires quite some manual steps anyway, I think updating these links as well shouldn't be a big deal. With explicit version links, you have the (small) benefit that even if people look would look at old versions of the documentation, the links would still point to the correct javadoc pages, while a `/…/latest|stable/…` link might break if at some point a class or package gets renamed (e.g. in #241).

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/object/OSMContribution.java
tyrasd: Perhaps a bit, but I thought this way is more logical: is it even a contribution/snapshot of the same OSM object -> is it about the same version -> is it about the same timestamp…

That said, I'm hesitating to merge this `equals` method as is. I've added it because after implementing `compareTo`, the code analysis rightly [complained](https://jenkins.ohsome.org/job/oshdb/job/documentation-groupByEntity/3/spotbugs/new/fileName.764779392/source.112c6af9-d6a5-4079-adcf-30f5540ce071/#272) that the class should also implement `equals` and it to be consistent with `compareTo` (i.e. such that `a.equals(b)` is true if and only if `a.compareTo(b) == 0`). This wasn't the case since the default `equals` method in java compares object references (`a == b`). But after implementing `equals` to match `compareTo`, we're now in a situation where (for example) two snapshots of the same entity at the same timestamp but with differing geometries would be considered _equal_, while they are not. Adding a equality check for the geometry could solve this, but would be super slow because it would have to calculate the objects' geometries first. Similarly, I also have some hard time finding a suitable way to implement `hashCode` (which should be consistent with `equals` and `compareTo` also).

I'm leaning toward removing the above implementation of `equals` while leaving the `compareTo` method as is and adding a javadoc comment to state that it doesn't match the default `equals`, perhaps by adding: _“Note: this class has a natural ordering that is inconsistent with equals."_ :shrug:

SlowMo24: but can that possibly happen: an object at one timestamp con only have one shape (in our database). It could only happen if you wanted to compare e.g. two geometry creation algorithms but in that case you would explicitly compare the geometry, right. 

The only sideeffect I could think of is that if someone (by error) specifies the same timestamp twice in a snapshot view she would get the same snapshot twice for the 'normal' mapReduce but only once when  `groupByEntity`.

tyrasd: Yes it can happen, because the `.getGeometry` returns the _clipped_ geometry of an entity at the specific timestamp. If you use (for example) the `.aggregateByGeometry` feature, then an OSM feature lying on the boundary of two "aggregate-by-polygons" will be returned as two snapshots/contribution objects where only the `.getGeometry` output differs.

joker234 (approved):
Do we need a Changelog comment on the compareability of `OSMContribution`?

tyrasd:
> Do we need a Changelog comment on the compareability of `OSMContribution`?

I guess it's not really needed since the change is quite subtle.

PR #251, opened 2020-06-25
fix doclint error (in Java 11)

tyrasd:
fixes an error while building javadoc on openjdk-11:

```
MavenReportException: Error while generating Javadoc:
[main] ERROR org.apache.maven.cli.MavenCli - Exit code: 1 - /var/jenkins_home/workspace/oshdb_master-RVIDC2LPT4HCVWT5BKWAIEUOKFCZO7TEKEF7P2PHZSCDMPY344BA/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java:19: error: attribute not supported in HTML5: summary
[main] ERROR org.apache.maven.cli.MavenCli - * <table summary="how XYGrid sees the world" style="text-align:center; border-spacing: 4px">
[main] ERROR org.apache.maven.cli.MavenCli - ^
```

### DIFF
--- a/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
+++ b/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
@@ -14,8 +14,9 @@ import org.slf4j.LoggerFactory;
 /**
  * XYGrid spans an equal degree grid over the world.
  * 
- * <p>How XYGrid sees the world. Example IDs for zoom = 2:
+ * <p>Example IDs for zoom = 2:
  * <table style="text-align:center; border-spacing: 4px">
+ * <caption>How XYGrid sees the world.</caption>
  * <tr>
  * <td></td><td></td><td colspan="3">+90 lat</td><td></td><td></td><td></td>
  * </tr>
--- a/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
+++ b/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
@@ -32,7 +32,7 @@ import org.slf4j.LoggerFactory;
  * <tr>
  * <td></td><td></td><td colspan="3">-90 lat</td><td></td><td></td><td></td>
  * </tr>
- * </table></p>
+ * </table>
  *
  * <p>Longitude +180 will be wrapped around to -180. Coordinates lying on
  * grid-borders will be assigned to the north-eastern cell.</p>
--- a/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
+++ b/oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
@@ -13,10 +13,9 @@ import org.slf4j.LoggerFactory;
 
 /**
  * XYGrid spans an equal degree grid over the world.
- * <br>
- * <br>
- * IDs for zoom = 2:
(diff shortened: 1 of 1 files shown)


oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
```diff
@@ -32,12 +32,10 @@
  * <tr>
  * <td></td><td></td><td colspan="3">-90 lat</td><td></td><td></td><td></td>
  * </tr>
- * </table>
- * <br>
- * Longitude +180 will be wrapped around to -180. Coordinates lying on
- * grid-borders will be assigned to the north-eastern cell.
- *
+ * </table></p>
```
joker234, line 35: Is `<table>` in `<p>` supposed to be right? I know it wasn't allowed in the past ([HTML 4](https://www.w3.org/TR/html401/struct/text.html#h-9.3.1)), but it might be allowed in [HTML 5](https://www.w3.org/TR/2018/WD-html53-20181018/grouping-content.html#the-p-element). I'm not really aware of the current rules there, the documentation says: _Some block elements cannot be children of p elements._

Jenkinsfile
```diff
@@ -47,7 +47,7 @@ pipeline {
           rtMaven.deployer.deployArtifacts = false
 
           withCredentials([string(credentialsId: 'gpg-signing-key-passphrase', variable: 'PASSPHRASE')]) {
-            buildInfo = rtMaven.run pom: 'pom.xml', goals: 'clean compile javadoc:jar source:jar install -P sign,git,withDep -Dmaven.repo.local=.m2 $MAVEN_TEST_OPTIONS -Dgpg.passphrase=$PASSPHRASE'
+            buildInfo = rtMaven.run pom: 'pom.xml', goals: 'clean compile javadoc:jar source:jar install -P sign,git,withDep -Dmaven.repo.local=.m2 $MAVEN_TEST_OPTIONS -Dgpg.passphrase=$PASSPHRASE -DdetectJavaApiLink=false'
```
joker234, line 50: ` -DdetectJavaApiLink=false` should be added to all `rtMaven.run` where `javadoc:…` is used. Maybe with a preceding comment that this should be removed in the future.

oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
tyrasd: AFAIK, javadoc requires one to put `<p>` after every empty line / before every paragraph. At least checkstyle complains otherwise:

> Empty line should be followed by <p> tag on the next line. (16:0) [JavadocParagraph]

Jenkinsfile
tyrasd: this was just a temporary "hotfix" to make it compile, so I can see jenkins' javadoc error messages. since the branch wouldn't trigger the deploy/publish-javadoc stages, it is enough to just do the hotfix here. Anyway, this wasn't supposed to get merged into master from this branch.

oshdb/src/main/java/org/heigit/bigspatialdata/oshdb/index/XYGrid.java
joker234: But javadoc also complains about the unnecessary `</p>` (compare [log](https://jenkins.ohsome.org/blue/organizations/jenkins/oshdb/detail/PR-251/3/pipeline#step-39-log-213)). I think you should setting the `</p>` before `<table>` would be the clean way to do it.

PR #258, opened 2020-08-11
catch all remote oshdb timeouts

tyrasd:
When a timeout happens remotely (in the ignite cluster), the resulting exception which is thrown in the client might be nested in various levels of `IgniteException`s (depending on where in the code they are thrown precisely). This implementation recursively unwinds these nested exceptions, so that the original `OSHDBTimeout` exception is actually thrown in all cases.

## Type of change

## Corresponding issue
Fixes https://github.com/GIScience/ohsome-api/issues/10

# Checklist
- [x] ~~I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) if necessary~~

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteAffinityCall.java
+++ b/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteAffinityCall.java
@@ -136,17 +136,21 @@ public class MapReducerIgniteAffinityCall<X> extends MapReducer<X>
     } catch (IgniteFutureTimeoutException e) {
       throw new OSHDBTimeoutException();
     } catch (IgniteException e) {
-      throw unwindIgniteExceptions(e);
+      // When a timeout happens remotely, the exception might be burried in (few) layers of
+      // "ignite exceptions". This recursively unwinds these and throws the original exception.
+      throw searchRemoteTimeout(e);
     }
   }
 
-  private static RuntimeException unwindIgniteExceptions(IgniteException exception) {
+  /** This recursively unwinds nested ignite exceptions: if an OSHDBTimeoutException is found it is
+   * returned, otherwise the causing exception is thrown.*/
+  private static OSHDBTimeoutException searchRemoteTimeout(IgniteException exception) {
     if (exception.getCause() == exception) {
       throw exception;
     } else if (exception.getCause() instanceof OSHDBTimeoutException) {
       return (OSHDBTimeoutException) exception.getCause();
     } else if (exception.getCause() instanceof IgniteException) {
-      return unwindIgniteExceptions((IgniteException) exception.getCause());
+      return searchRemoteTimeout((IgniteException) exception.getCause());
     } else {
       throw exception;
     }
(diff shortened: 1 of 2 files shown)


oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteAffinityCall.java
```diff
@@ -136,11 +136,23 @@ public boolean isActive() {
...
-      }
+      // When a timeout happens remotely, the exception might be burried in (few) layers of
+      // "ignite exceptions". This recursively unwinds these and throws the original exception.
+      throw searchRemoteTimeout(e);
+    }
+  }
+
+  /** This recursively unwinds nested ignite exceptions: if an OSHDBTimeoutException is found it is
+   * returned, otherwise the causing exception is thrown.*/
+  private static OSHDBTimeoutException searchRemoteTimeout(IgniteException exception) {
```
rtroilo, line 147: basicly what searchRemoteTimeout does is unwrapping the IgniteException(s) chain.
Why not just return the unwrapped cause of the IgniteException(s) 
like:
```java
Throwable unwrap(IgniteException e) {
  Throwable cause = e.getCause();
  if (cause instanceof IgniteException && e != cause) {
    return unwrap((IgniteException) cause);
  }
  return cause;
}
```
On the receiver side we can check for a specific cause if needed. In this pull request it is not.

we can even pull this function into an ignite-utility class.

rtroilo: `e.getCause()` could be null. [api-doc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Throwable.html#getCause())

so the receiver side should also check for null before re-throwing the cause.

tyrasd: > Why not just return the unwrapped cause of the IgniteException(s)

The current version is in my eyes preferable, becuase it is more easily comprehendable (because it is less generic), and we don't need the more flexible "utility" method. If we need one at some point in the future, I would like to keep a `searchRemoteTimeout` method which then would just call the new method and do the null-check, e.g.

```java
private static OSHDBTimeoutException searchRemoteTimeout(IgniteException exception) {
  Throwable unwrapped = ….unwrap(exception);
  if (unwrapped instanceof OSHDBTimeoutException) {
    return (OSHDBTimeoutException) unwrapped;
  } else {
    throw exception;
  }
}
```

rtroilo: I like the new version of searchRemoteTimeout using unwrap.
It is much easier to read what it does. :+1: 

But at the moment, as far as I saw it in this pull request, the OSHDBTimeoutException returned from searchRemoteTimeout is throw immediately! 
Do we have a use for the returned OSHDBTimeoutException somewhere else?

tyrasd: > Do we have a use for the returned OSHDBTimeoutException somewhere else?

Don't think so. Do you think it would be nicer to write it like this then:

```java
private static OSHDBTimeoutException searchRemoteTimeout(IgniteException exception) {
  Throwable unwrapped = ….unwrap(exception);
  if (unwrapped instanceof OSHDBTimeoutException) {
    throw (OSHDBTimeoutException) unwrapped;
  } else {
    throw exception;
  }
}
```

tyrasd: I've changed this now as in the last suggestion, but inlined the `searchRemoteTimeout` method.

Issue #90, opened 2019-03-07
Solve Checkstyle Warnings


SlowMo24:
As of now, 6533 CheckStyle Warnings are recorded. These are no code errors but rather coding style errors. This is a static analyses and issues may not be solved or be irrelevant.

The goal is to solve solve as may of these as possible and close the issue. After that, increases in reports can be handled e.g on each merge into master.

The following modules need checking:
- [ ] etl (2240)
- [ ] oshdb (919)
- [x] oshdb-api -> #279
- [x] oshdb-util -> #275, #283
- [x] ~oshpbf-parser~ (will be removed, doesn't need to be fixed)

State: [2020-12-07, build master/367](https://jenkins.ohsome.org/job/oshdb/job/master/367/checkstyle/#moduleNameContent)

PR #283, opened 2020-11-12
fix checkstyle issues and other in oshdb-util

joker234:
## Type of change
- [x] Code quality improvement

## Description
- Fixes newly introduced checkstyle issues in oshdb-util
- Refactor oshdb-util tests

## Corresponding issue
Supports #90, related to 91acd97725feec2da5d4e2de18d111a07593ce44

# 🚧 Checklist
### Additional checklist/🚧 items
- [x] fix newly introduced checkstyle issues
- [x] refactor tests
- [x] Decide about unparameterized tests:
  - Resolve: _Replace these 3 tests with a single Parameterized one._ https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmhistorytestdata/OSHDBGeometryBuilderTestOsmHistoryTestDataRelationNotMultipolygonTest.java#L220
  - Resolve: _Replace these 5 tests with a single Parameterized one._ https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData1xxTest.java#L167
  - Resolve: _Replace these 4 tests with a single Parameterized one._ https://github.com/GIScience/oshdb/blob/d201f3837dfac3c17c9b1ffcb52314c46c7605d5/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData3xxTest.java#L40
  - Resolve: _Replace these 15 tests with a single Parameterized one._ https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java#L508
  - Resolve: _Replace these 3 tests with a single Parameterized one._ https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java#L608
  - Resolve: _Replace these 3 tests with a single Parameterized one._ https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java#L1164
  - Resolve: _Replace these 4 tests with a single Parameterized one._ https://github.com/GIScience/oshdb/blob/d201f3837dfac3c17c9b1ffcb52314c46c7605d5/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/relations/OSHDBGeometryBuilderRelationOuterDirectionsTest.java#L34
- [x] Solve issues in `OSMXmlReader.java`
- [x] Resolve questions (maybe @tyrasd wants to take a look at them):
  - [x] Why has `GridOSHFactory.java` two `getGridOSHWays`, although, `cellZoom` and `cellId` are never used? https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/helpers/GridOSHFactory.java#L68 https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/helpers/GridOSHFactory.java#L81
  - [x] Same question for `getGridOSHRelations`.
  - [x] Does it make sense to refactor the many asserts in two tests, as IDEA suggests in `FastBboxInPolygonTest.java`?
  - [x] Does it make sense to refactor the many asserts in two tests, as IDEA suggests in `FastBboxOutsidePolygonTest.java`?
  - [x] What is the need or requirement for the many try-catch blocks that just throw the exception again in tests? e.g. https://github.com/GIScience/oshdb/blob/c637936209abe4834f751e4f16cec350ae49e7fc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmhistorytestdata/OSHDBGeometryBuilderTestOsmHistoryTestDataRelationNotMultipolygonTest.java#L473-L483
  - [x] What should this do? This test is always true: https://github.com/GIScience/oshdb/blob/d201f3837dfac3c17c9b1ffcb52314c46c7605d5/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData3xxTest.java#L35-L36
  - [x] What to do with these empty Tests? In the current solution, they just get commented out. E.g. https://github.com/GIScience/oshdb/blob/d201f3837dfac3c17c9b1ffcb52314c46c7605d5/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData3xxTest.java#L51-L55
  - [x] Is the class `FakeTagInterpreter` still necessary? https://github.com/GIScience/oshdb/blob/d201f3837dfac3c17c9b1ffcb52314c46c7605d5/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/OSHDBGeometryBuilderTest.java#L221-L238
  - [x] Is the method `MutableOSMEntity.setEntity(…)` still necessary? https://github.com/GIScience/oshdb/blob/b1548dec2e9743e0c6de57ad4fef7d4dd08d3edc/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/xmlreader/MutableOSMEntity.java#L19

### DIFF
--- a/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/IterateByContributionRelationsTest.java
+++ b/oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/IterateByContributionRelationsTest.java
@@ -37,10 +37,13 @@ public class IterateByContributionRelationsTest {
    * {@link GridOSHRelations}.
    */
   public IterateByContributionRelationsTest() throws IOException {
-    osmXmlTestData.add("./src/test/resources/different-timestamps/polygon.osm");// read osm xml data
-    areaDecider = new OSMXmlReaderTagInterpreter(osmXmlTestData);// Used to provided information
+    // read osm xml data
+    osmXmlTestData.add("./src/test/resources/different-timestamps/polygon.osm");
+    // Used to provided information
+    areaDecider = new OSMXmlReaderTagInterpreter(osmXmlTestData);
     // needed to create actual geometries from OSM data
-    oshdbDataGridCell = GridOSHFactory.getGridOSHRelations(osmXmlTestData);// get GridOSH's
+    // get GridOSH's
+    oshdbDataGridCell = GridOSHFactory.getGridOSHRelations(osmXmlTestData);
     // (Holds the basic information, every OSM-Object has at a specific level) out of osm-xml file
   }
 
@@ -55,12 +58,12 @@ public class IterateByContributionRelationsTest {
             "2020-01-01T00:00:00Z"
         ).get(),
         // look at dat in this bbox
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         // needed to create actual geometries from OSM data
         areaDecider,
         // oshEntityPreFilter: get data of relation with id 500
         oshEntity -> oshEntity.getId() == 500,
-        osmEntity -> true,// osmEntityFilter: true -> get all
+        osmEntity -> true, // osmEntityFilter: true -> get all
         false
     )).iterateByContribution(
         oshdbDataGridCell
@@ -100,7 +103,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 501,
         osmEntity -> true,
@@ -134,7 +137,7 @@ public class IterateByContributionRelationsTest {
               "2000-01-01T00:00:00Z",
               "2020-01-01T00:00:00Z"
           ).get(),
-          new OSHDBBoundingBox(-180,-90, 180, 90),
+          new OSHDBBoundingBox(-180, -90, 180, 90),
           areaDecider,
           oshEntity -> oshEntity.getId() == 502,
           osmEntity -> true,
@@ -156,7 +159,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 503,
         osmEntity -> true,
@@ -190,7 +193,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 504,
         osmEntity -> true,
@@ -230,7 +233,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 505,
         osmEntity -> true,
@@ -269,7 +272,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 506,
         osmEntity -> true,
@@ -308,7 +311,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 507,
         osmEntity -> true,
@@ -342,7 +345,7 @@ public class IterateByContributionRelationsTest {
               "2000-01-01T00:00:00Z",
               "2020-01-01T00:00:00Z"
           ).get(),
-          new OSHDBBoundingBox(-180,-90, 180, 90),
+          new OSHDBBoundingBox(-180, -90, 180, 90),
           areaDecider,
           oshEntity -> oshEntity.getId() == 508,
           osmEntity -> true,
@@ -364,7 +367,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 509,
         osmEntity -> true,
@@ -421,7 +424,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 510,
         osmEntity -> true,
@@ -448,7 +451,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 511,
         osmEntity -> true,
@@ -485,7 +488,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 512,
         osmEntity -> true,
@@ -519,7 +522,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 513,
         osmEntity -> true,
@@ -549,7 +552,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 514,
         osmEntity -> true,
@@ -584,7 +587,7 @@ public class IterateByContributionRelationsTest {
             "2000-01-01T00:00:00Z",
             "2020-01-01T00:00:00Z"
         ).get(),
-        new OSHDBBoundingBox(-180,-90, 180, 90),
+        new OSHDBBoundingBox(-180, -90, 180, 90),
         areaDecider,
         oshEntity -> oshEntity.getId() == 515,
         osmEntity -> true,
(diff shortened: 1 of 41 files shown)


tyrasd:
> Why has GridOSHFactory.java two getGridOSHWays, although, cellZoom and cellId are never used?
> Same question for getGridOSHRelations.

The parameters should have been passed through to `GridOSHWays.compact` and `GridOSHRelations.compact`. It wasn't a problem since the tests (as of now) apparently don't care about the returned cell's id and zoom level. But that might change. I'll fix it.

> Does it make sense to refactor the many asserts in two tests, as IDEA suggests in FastBboxInPolygonTest.java?
> Does it make sense to refactor the many asserts in two tests, as IDEA suggests in FastBboxOutsidePolygonTest.java?

If you ask me they can stay like this. In order to fully test the method on a polygon one simply has to test a few possible cases how a bbox can interact with the polygon :shrug: 

Btw: these are suggestions by _SonarLint_ and not from IDEA directly, right? It says: “_A common good practice is to write test methods targeting only one logical concept, that can only fail for one reason. While it might make sense to have more than one assertion to test one concept, having too many is a sign that a test became too complex and should be refactored to multiples ones._”

I've added `@SuppressWarnings("java:S5961")` in these places, what do you think?

> What is the need or requirement for the many try-catch blocks that just throw the exception again in tests?

As far as I remember, for some of the test cases we said that we really don't care about how exactly a certain corner case situation is handled as long as the code does not crash with an exception (e.g. when the `uid field is missing, then we don't care if it is filled with `0`, `-1` or any other imaginable value). 

But if you just write a test without the try-catch->fail statement, tools like SonarLint complain that each test should have at least one assert statement. Maybe there's a better way to write such a "relaxed" test? :thinking: adding a `@SuppressWarnings("java:S2699")` can silence the warnings, but adding `@Test(expected = Test.None.class /* no exception expected */)` (see https://stackoverflow.com/a/44801486/1627467) might be more explicit way to state this (even though IntelliJ then suggests to remove the default parameter assignment :shrug: ).

> What should this do? This test is always true:

Looks like a mistake in the test. It should actually test the value of the stored `uid` like this: 

```java
    int entityUid = testData.nodes().get(200000L).get(0).getUserId();
    assertEquals(1, entityUid);
```

> What to do with these empty Tests? In the current solution, they just get commented out. E.g.

There are not really that useful for us because they test cases we don't really have to deal with (at least at the moment). But I'll replace them with something slightly more useful anyway.

> Is the class FakeTagInterpreter still necessary?

Nope, not this one. It was replaced by the specific classes like `FakeTagInterpreterAreaNever`. It's removed.

> Is the method MutableOSMEntity.setEntity(…) still necessary?

I guess not. Removed now.

tyrasd:
> What to do with these empty Tests? 

I realized that there are a few more places where there are still empty tests. I guess it would be best to look through them one by one and if they are really not needed just replace them with a single comment, for a hypothetical example

```java
  @Test
  public void test666() {
    // 666: a whole bunch of nodes!
    // the same like test100() just with more entities
  }
```

could become

```java
  // test666 (a whole bunch of nodes!) skipped because it's the same as the same test100() just with more entities
```

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/incomplete/OSHDBGeometryBuilderTestPolygonIncompleteDataTest.java
```diff
@@ -69,6 +69,7 @@ public void testWayNotExistent() throws ParseException {
     assertEquals(expectedPolygon.getArea(), intersection.getArea(), DELTA);
   }
 
+  @SuppressWarnings({"UnusedAssignment", "unused"})
```
tyrasd, line 72: maybe it would be better to get rid of the unused variable/assignment?

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmhistorytestdata/OSHDBGeometryBuilderTestOsmHistoryTestDataRelationTest.java
```diff
@@ -110,6 +110,7 @@ public void testVisibleChange() throws ParseException {
     assertEquals(expectedPolygon2.getArea(), intersection2.getArea(), DELTA);
   }
 
+  @SuppressWarnings({"unused", "UnusedAssignment"})
```
tyrasd, line 113: or instead simply drop variable `result`?

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmhistorytestdata/OSHDBGeometryBuilderTestOsmHistoryTestDataRelationTest.java
```diff
@@ -275,14 +276,15 @@ public void testGeometryCollection() {
...
-      assertTrue(result.getNumGeometries() == 6);
+      assertEquals(6, result.getNumGeometries());
       assertFalse(result instanceof MultiPolygon);
     } catch (Exception e) {
       e.printStackTrace();
       fail("Should not have thrown any exception");
     }
   }
 
+  @SuppressWarnings({"unused", "UnusedAssignment"})
```
tyrasd, line 287: see above

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData1xxTest.java
```diff
@@ -233,14 +233,15 @@ public void test130() {
...
-      for (int i = 0; i < result2.getLength();i++) {
+    for (int j = 0; j < result1.getLength(); j++) {
+      for (int i = 0; i < result2.getLength(); i++) {
         assertNotEquals(((LineString) result1).getCoordinateN(j),
             ((LineString) result2).getCoordinateN(i));
       }
     }
   }
 
+  @SuppressWarnings("ResultOfMethodCallIgnored")
```
tyrasd, line 244: the logic of the final part of this test is flawed. I think what the author tried to do is something like:

```java
    boolean hasCommonCoordinate = false;
    for (int j = 0; j < result1.getLength(); j++) {
      for (int i = 0; i < result2.getLength(); i++) {
        Coordinate c1 = ((LineString) result1).getCoordinateN(j);
        Coordinate c2 = ((LineString) result2).getCoordinateN(i);
        if (c1.equals(c2)) {
          hasCommonCoordinate = true;
        }
      }
    }
    assertTrue(hasCommonCoordinate);
```

:thinking:

tyrasd: or slightly shorter:

```java
    Set<Coordinate> res1Coords = Arrays.stream(result1.getCoordinates()).collect(Collectors.toSet());
    assertTrue(Arrays.stream(result1.getCoordinates()).anyMatch(res1Coords::contains));
```

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData1xxTest.java
```diff
@@ -263,6 +264,7 @@ public void test131() {
...
```
tyrasd, line 267: same as above

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/IterateByContributionRelationsTest.java
```diff
@@ -37,10 +37,13 @@
    * {@link GridOSHRelations}.
    */
   public IterateByContributionRelationsTest() throws IOException {
-    osmXmlTestData.add("./src/test/resources/different-timestamps/polygon.osm");// read osm xml data
-    areaDecider = new OSMXmlReaderTagInterpreter(osmXmlTestData);// Used to provided information
+    // read osm xml data
+    osmXmlTestData.add("./src/test/resources/different-timestamps/polygon.osm");
+    // Used to provided information
+    areaDecider = new OSMXmlReaderTagInterpreter(osmXmlTestData);
     // needed to create actual geometries from OSM data
```
tyrasd, line 44: :thinking: I guess this comment should be appended to the comment which was previously after the `;` at the end of the line above:

> Used to provide information needed to create actual geometries from OSM data

(also there seems to be a typo in there: _provided_ -> _provide_)

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/IterateByContributionRelationsTest.java
```diff
@@ -37,10 +37,13 @@
...
-    oshdbDataGridCell = GridOSHFactory.getGridOSHRelations(osmXmlTestData);// get GridOSH's
+    // get GridOSH's
+    oshdbDataGridCell = GridOSHFactory.getGridOSHRelations(osmXmlTestData);
     // (Holds the basic information, every OSM-Object has at a specific level) out of osm-xml file
```
tyrasd, line 47: this line also belongs to the one two lines above (L45):

> get GridOSH's (Holds the basic information, every OSM-Object has at a specific level) out of osm-xml file

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/IterateByTimestampsRelationsTest.java
```diff
@@ -92,6 +92,7 @@ public void testVisibleChange() {
     assertEquals(303, result.get(0).osmEntity.getChangesetId());
   }
 
+  @SuppressWarnings("unused")
```
tyrasd, line 95: maybe it's better to use `ResultOfMethodCallIgnored` suppression here and remove the unused local variable?

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/IterateByTimestampsRelationsTest.java
```diff
@@ -254,6 +255,7 @@ public void testGeometryCollection() {
     assertTrue(geom2 instanceof GeometryCollection);
   }
 
+  @SuppressWarnings("unused")
```
tyrasd, line 258: same as above

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java
```diff
@@ -1160,6 +1160,7 @@ public void test791() {
     }
   }
 
+  @SuppressWarnings("unused")
```
tyrasd, line 1163: maybe better drop the variable then? also below (`test793` and `795`)

joker234:
Decided to not replace the _Replace these x tests with a single Parameterized one._ issues.

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/celliterator/IterateByTimestampsRelationsTest.java
joker234: `ResultOfMethodCallIgnored` is incorrect. The Problem is, that the assigned variable `result` is never used.

joker234: removed the assignment.

oshdb-util/src/test/java/org/heigit/bigspatialdata/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData1xxTest.java
joker234: What does this actually do? I read it as `res1Coords` is a set of `result1`s coordinates. And then the coordinates of `result1` are checked if they are inside of this set :raised_eyebrow: :thinking:

tyrasd: oops, yes, the second line should be with `result2` to actually make sense. you version is what I had in mind. :bow:

joker234: I think 1705e2cb fixed this.

tyrasd (approved):
nice improvements :+1:

Issue #302, opened 2020-12-08
integrate "ohsome filter" library (as oshdb-filter module), deprecate "old" filters


tyrasd:
see corresponding [gitlab issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter/-/issues/15).

* [x] import code as new module (`oshdb-filter`) in this repository
* [x] deprecate `osmTag`, `osmType` filters
* [x] update documentation, examples, benchmarks
  * [x] rename ohsome filter to oshdb filter https://github.com/GIScience/oshdb/blob/66c04e5f1d440d2e79a393682ec2b9354d62319c/documentation/manual/filters.md#L47
  * [x] replace GitLab link https://github.com/GIScience/oshdb/blob/66c04e5f1d440d2e79a393682ec2b9354d62319c/documentation/manual/filters.md#L50

joker234:
I think it's easier to merge #304 first.

PR #306, opened 2020-12-08
integrate ohsome filter

tyrasd:
see #302 

## New or changed dependencies:
- dropped filter library, integrated here directly

# Checklist
- ~I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/master/documentation)~ see https://github.com/GIScience/oshdb/pull/306#issuecomment-741826839
- ~I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) if necessary~ see https://github.com/GIScience/oshdb/pull/306#issuecomment-741826839


tyrasd:
documentation, examples (and benchmarks?) will be updated in a separate PR

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+## 0.7.0-SNAPSHOT (current master)
+
+* integrate [ohsome-filter](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter) module fully into this repository, renaming it to `oshdb-filter`. #306
```
joker234, line 6: Would there be transition documentation about how to further use the filter functionality? Like when changing the oshdb groupId:
> When switching to the OSHDB version 0.6 you need to adapt your `pom.xml` to the new groupId, e.g.:
> ```xml
> <dependency>
>   <groupId>org.heigit.ohsome</groupId>
>   <artifactId>oshdb-api</artifactId>
>   <version>0.6.0</version>
> </dependency>
> ```

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -275,7 +276,9 @@ private MapAggregator(MapAggregator<U, ?> obj, MapReducer<IndexValuePair<U, X>>
    *
    * @param typeFilter the set of osm types to filter (e.g. `EnumSet.of(OSMType.WAY)`)
    * @return a modified copy of this object (can be used to chain multiple commands together)
+   * @deprecated use oshdb-filter {@link #filter(String)} instead
```
joker234, line 279: Are these `deprecated` changes because of the integration of ohsome filter? Or are these deprecations a general recommendation?

oshdb-filter/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+A parser and interpreter for filters which can be applied on OSM entities. It allows, for example, to filter by various combinations of OSM tags.
+
+Before this was integrated in the main OSHDB repository, this was developed as an independent library called [“_ohsome filter_”](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter).
+
+Usage
+-----
+
+Filters can be used directly in the OSHDB:
+
```
joker234, line 12: I think somewhere has to be mentioned that you should add the dependency. This was here in an older version of the README.

oshdb-filter/README.md
```diff
@@ -0,0 +1,108 @@
...
+| highways | `type:way and highway in (motorway, motorway_link, trunk, trunk_link, primary, primary_link, secondary, secondary_link, tertiary, tertiary_link, unclassified, residential, living_street, pedestrian) or (highway=service and service=alley))` | The list of used tags depends on the exact definition of a "highway". In a different context, it may also incude less or even more highway tags (like `footway`, `cycleway`, `track`, `path`, all `highway=service`, etc.). |
+| residential roads missing a name (for quality assurance) | `type:way and highway=residential and name!=* and noname!=yes` | Note that some roads might be actually unnamed in reality. Such features can be marked as unnamed with the [`noname`](https://wiki.openstreetmap.org/wiki/Key:noname) tag in OSM. |
+| buildings with implausibly big footprints | `geometry:polygon and building=* and building!=no and area:(1E6..)` | The currently largest building by footprint area is a [car factory measuring about 887,800 m²](https://en.wikipedia.org/wiki/List_of_largest_buildings#Largest_footprint). |
+
```
joker234, line 108: Here was a Reference to the javadoc. Do we need another reference here or is the general reference sufficient?

oshdb-filter/pom.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.heigit.ohsome</groupId>
+    <artifactId>oshdb-parent</artifactId>
+    <version>0.7.0-SNAPSHOT</version>
```
joker234, line 10: General question: can't we use `<version>${project.version}</version>` here?

joker234 (changes requested):
Open ToDo:
- [x] fix `"ohsome filter"` https://github.com/GIScience/oshdb/blob/0f504158d9c2fd1c2dc2c09903b06ea7b4402320/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java#L2073
- [x] missing `@Deprecated`? https://github.com/GIScience/oshdb/blob/0f504158d9c2fd1c2dc2c09903b06ea7b4402320/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java#L371
- [ ] missing `@Deprecated`? https://github.com/GIScience/oshdb/blob/0f504158d9c2fd1c2dc2c09903b06ea7b4402320/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducer.java#L599
- [ ] change deprecation reference to new filter for all `where(…)` functions? This function and the following: https://github.com/GIScience/oshdb/blob/0f504158d9c2fd1c2dc2c09903b06ea7b4402320/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapReducerSettings.java#L149

joker234:
We don't want to keep the ohsome-filter Changelog, correct? Removed in c15e945bd5c41e08f3041e0e71914a9826c88320

tyrasd:
> We don't want to keep the ohsome-filter Changelog, correct? Removed in [c15e945](https://github.com/GIScience/oshdb/commit/c15e945bd5c41e08f3041e0e71914a9826c88320)

not really (it's still available in the original repository for reference). thanks :+1:

joker234:
General comment: The review only included all files outside of `/oshdb-filter` and all changes in `/oshdb-filter` between 633bdcb2 and 0f504158d9c2fd1c2dc2c09903b06ea7b4402320 (current `integrate-ohsome-filter`). The changes before this commit are not reviewed as they were part of the old repository already.

tyrasd:
> missing @Deprecated? [x2]

no, because the oshdb-filter doesn't have an equivalent for the `String key, Pattern valuePattern` case at the moment. see https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter/-/issues/11

tyrasd:
> change deprecation reference to new filter for all where(…) functions?

ui, I totally forgot these still exist. I think it's time to drop them with 0.7, but let's do that in a separate PR

CHANGELOG.md
tyrasd: true, let me add that

tyrasd: well… actually, nothing much changed. Only when one used the `filter(FilterExpression)` method (e.g. via `.filter(Filter.byOSHEntity(…))` one would need to update the package name in the import. Otherwise everything would stay the same. Or am I overlooking something here?

oshdb-filter/pom.xml
joker234: Just checked that myself. Properties are prohibited in parent definitions.

CHANGELOG.md
joker234: I think you have better insight. It was more meant like a question if you want to add something like that here or not.

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
tyrasd: well, I would say that since the oshdb-filter is a more capable interface there's no need to drag around dozens of overloaded methods just for convenience.

But that's in principle up for discussion. In some situations it might be simpler to write `osmTag("key with spaces")` compared to `filter("\"key with spaces\" = *")` (note the quoting required by the spaces present in the tag key), but there are more like rare corner cases IMHO.

oshdb-filter/README.md
tyrasd: I guess it should be fine like this. the reference was just here from when the filter was it's own repo.

oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/MapAggregator.java
joker234: The question was more meant if this deprecation is dependent on the *integration* of ohsome-filter as oshdb-filter, or if this could have been done anyway.

joker234 (approved):

PR #319, opened 2020-12-18
maxzoom is 14 not 15

rtroilo:
hard coded maxzoom is actual 14 not 15
## Type of change
change hard coded maxzoom from 15 to 14

## Description
currently the OSHDB.MAXZOOM is used by 
- MapReducer.getCellIdRanges

In the current and the last past, datasets the maxzoom was/is actually 14 and so this function does more work then actually needed.


tyrasd (approved):
oh wow :see_no_evil:

PR #320, opened 2020-12-18
🚧 use int for internal coordinate representation and add static function to convert coordinates to int/double and vice versa

rtroilo:
improve conversion from long coordinate to double coordinate, 
## Type of change

## Description
the current implementation leads to a "not so nice" representation of long value coordinate as a double coordinate
```
 long l = 1231234567L; // internal long for 123.1234567;
 double d = l * OSHDB.GEOM_PRECISION;
 -> 123.12345669999999
```

This pull request adds static functions to the OSHDB class
- coordinateToLong, coordinateToDouble
which improve the conversion from long value coordinates to double value and vice versa.

```
  long l = 1231234567L;
  double convertedDouble = OSHDB.coordinateToDouble(l);
  assertEquals(123.1234567, convertedDouble,0.0);
```    
 
# Checklist
- [ ] My code follows the [code-style](https://github.com/GIScience/oshdb/blob/master/CONTRIBUTING.md) rules and I have checked on the [static analyses](https://jenkins.ohsome.org/job/oshdb/view/change-requests/) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) (if applicable) results
- [ ] I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/master/documentation)
- [ ] I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) if necessary


oshdb-tool/etl/src/main/java/org/heigit/ohsome/oshdb/tool/importer/util/ZGrid.java
```diff
@@ -1,5 +1,6 @@
 package org.heigit.ohsome.oshdb.tool.importer.util;
 
+import static org.heigit.ohsome.oshdb.OSHDB.coordinateToLong;
```
tyrasd, line 3: there should be an empty line here

* https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing
* https://jenkins.ohsome.org/job/oshdb/view/change-requests/job/PR-320/2/checkstyle/fileName.1640957702/source.f0ea342c-43cd-46a2-aa73-1eede0ab77f9/#4

tyrasd, line 3: there should be an empty line here

* https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing
* https://jenkins.ohsome.org/job/oshdb/view/change-requests/job/PR-320/2/checkstyle/fileName.1640957702/source.f0ea342c-43cd-46a2-aa73-1eede0ab77f9/#4

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDB.java
```diff
@@ -8,6 +8,26 @@
   public static final long GEOM_PRECISION_TO_LONG = 10000000L;
   public static final double GEOM_PRECISION = 1.0 / GEOM_PRECISION_TO_LONG;
 
+  /**
+   * Convert a double precision coordinate to fix precision long value
```
tyrasd, line 12: Missing space after javadoc comment: https://jenkins.ohsome.org/job/oshdb/view/change-requests/job/PR-320/2/checkstyle/fileName.-449784562/type.-510536328/source.e4b60da5-07ad-4f97-b28c-1e166328b54c/#11 (same below)

tyrasd: is there a typo? “_…to fix**ed** precision long value_” would be better?! (same below also)

tyrasd:
```suggestion
   * Converts a double precision coordinate to a fixed precision long value.
```

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDB.java
```diff
@@ -8,6 +8,26 @@
...
+   * 
+   * @param coordinate
+   * @return converted coordinate
+   */
+  public static long coordinateToLong(double coordinate) {
+    return (long) (coordinate * GEOM_PRECISION_TO_LONG);
+  }
+
+  /**
+   * Convert a fix precision long coordinate to a double value
```
tyrasd, line 22: see above, but maybe something like this would be more readable:

```suggestion
   * Returns the double precision coordinate of a fixed precision long value.
```

oshdb/src/main/java/org/heigit/ohsome/oshdb/index/XYGrid.java
```diff
@@ -1,5 +1,6 @@
 package org.heigit.ohsome.oshdb.index;
 
+import static org.heigit.ohsome.oshdb.OSHDB.coordinateToLong;
```
tyrasd, line 3:
```suggestion
import static org.heigit.ohsome.oshdb.OSHDB.coordinateToLong;

```

oshdb-tool/etl/src/main/java/org/heigit/ohsome/oshdb/tool/importer/util/ZGrid.java
tyrasd:
```suggestion
import static org.heigit.ohsome.oshdb.OSHDB.coordinateToLong;

```

oshdb/src/main/java/org/heigit/ohsome/oshdb/index/XYGridTree.java
```diff
@@ -1,5 +1,6 @@
 package org.heigit.ohsome.oshdb.index;
 
+import static org.heigit.ohsome.oshdb.OSHDB.coordinateToLong;
```
tyrasd, line 3:
```suggestion
import static org.heigit.ohsome.oshdb.OSHDB.coordinateToLong;

```

oshdb/src/main/java/org/heigit/ohsome/oshdb/io/OSHDBByteArrayOutputStream.java
```diff
@@ -0,0 +1,58 @@
+package org.heigit.ohsome.oshdb.io;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Arrays;
+
+public class OSHDBByteArrayOutputStream extends OutputStream {
```
tyrasd, line 7: is this class even used in this PR, or did it sneak in by mistake?

oshdb/src/main/java/org/heigit/ohsome/oshdb/io/OSHDBInput.java
```diff
@@ -0,0 +1,243 @@
+package org.heigit.ohsome.oshdb.io;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.NoSuchElementException;
+import java.util.PrimitiveIterator.OfInt;
+import java.util.PrimitiveIterator.OfLong;
+
+public abstract class OSHDBInput {
```
tyrasd, line 9: is this class even used in this PR, or did it sneak in by mistake?

oshdb/src/main/java/org/heigit/ohsome/oshdb/io/OSHDBOutput.java
```diff
@@ -0,0 +1,80 @@
+package org.heigit.ohsome.oshdb.io;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class OSHDBOutput extends OutputStream {
```
tyrasd, line 6: is this class even used in this PR, or did it sneak in by mistake?

oshdb/src/main/java/org/heigit/ohsome/oshdb/osm/OSMNode.java
```diff
@@ -1,8 +1,8 @@
 package org.heigit.ohsome.oshdb.osm;
 
+import static org.heigit.ohsome.oshdb.OSHDB.coordinateToDouble;
```
tyrasd, line 3:
```suggestion
import static org.heigit.ohsome.oshdb.OSHDB.coordinateToDouble;

```

oshdb/src/main/java/org/heigit/ohsome/oshdb/utils/ByteUtils.java
```diff
@@ -0,0 +1,49 @@
+package org.heigit.ohsome.oshdb.utils;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+
+public abstract class ByteUtils {
```
tyrasd, line 10: should this class be a part of this PR?

oshdb/src/main/java/org/heigit/ohsome/oshdb/utils/OSHDBTags.java
```diff
@@ -0,0 +1,17 @@
+package org.heigit.ohsome.oshdb.utils;
+
+public abstract class OSHDBTags {
```
tyrasd, line 3: should this class be a part of this PR?

oshdb/src/test/java/org/heigit/ohsome/oshdb/OSHDBTest.java
```diff
@@ -0,0 +1,19 @@
+package org.heigit.ohsome.oshdb;
+
+import static org.junit.Assert.*;
```
tyrasd, line 3:
```suggestion
import static org.junit.Assert.*;

```

tyrasd (changes requested):
a few general questions:

1. why classify this a `bug`?
2. does it make sense that these methods live in the `OSHDB` class? at first glance, it seems like that place looks a bit too prominent?!
3. basically what the implementation does to avoid rounding errors is to use a floating point division instead of a multiplication. note that divisions are not as fast as multiplications on typical CPUs, so this might come with a small performance overhead. I'd guess that for us it wouldn't make a huge difference, but it would perhaps be nice to verify that by some benchmarking.

rtroilo:
> 2. does it make sense that these methods live in the `OSHDB` class? at first glance, it seems like that place looks a bit too prominent?!

According to your question, it makes no sense for you, right? So where do you want to have this methods then?

> 3. basically what the implementation does to avoid rounding errors is to use a floating point division instead of a multiplication. note that divisions are not as fast as multiplications on typical CPUs, so this might come with a small(?) performance overhead. I'd guess that for us it wouldn't make a huge difference, but it would perhaps be nice to verify that by some benchmarking.

You are right, the multiplication introduces a slightly worse performance. In this case I undo the conversion, to a multiplication over division. The now introduce rounding errors should be handled in outputing methods e.g.
```java
  System.out.printf("%.7f, %.7f", lon, lat);
``` 
  or for geometries with a precision model
```java
  Point point = ...;
  var pm = new PrecisionModel(10000000);
  var writer = new WKTWriter();
  writer.setPrecisionModel(pm);
  writer.write(point);
```

tyrasd:
> > does it make sense that these methods live in the `OSHDB` class? at first glance, it seems like that place looks a bit too prominent?!
> 
> According to your question, it makes no sense for you, right? So where do you want to have this methods then?

If you ask me: I'd put it somewhere near where they belong conceptionally. And since the coordinates belong (only) to the `OSMNode`s it might make sense to add it either to the `OSMNode` class, or perhaps in a new utility class (`OSMNodes`?).

rtroilo:
> > > does it make sense that these methods live in the `OSHDB` class? at first glance, it seems like that place looks a bit too prominent?!
> > 
> > 
> > According to your question, it makes no sense for you, right? So where do you want to have this methods then?
> 
> If you ask me: I'd put it somewhere near where they belong conceptionally. And since the coordinates belong (only) to the `OSMNode`s it might make sense to add it either to the `OSMNode` class, or perhaps in a new utility class (`OSMNodes`?).

OSHDBBoundigBox also use internally the long coordinate system, and do the long to double conversation.

tyrasd:
> > I'd put it somewhere near where they belong conceptionally. And since the coordinates belong (only) to the `OSMNode`s it might make sense to add it either to the `OSMNode` class, or perhaps in a new utility class (`OSMNodes`?).
> 
> OSHDBBoundingBox also use internally the long coordinate system, and do the long to double conversation.

True. Then maybe call the utility class `Coordinates`, `EncodedCoordinates` or something similar? Do you have any suggestion?

rtroilo:
sorry my last push was not verified!

CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ import org.heigit.ohsome.oshdb.api.db.OSHDBH2;
 * integrate [ohsome-filter](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter) module fully into this repository, renaming it to `oshdb-filter`. ([#306])
 * make sure predicate-filters are always serializable. ([#353])
 * improve maintainability of parts of important central processing algorithms for determining entity modification history: refactoring improves code structure, adds inline documentation and enhances test coverage. ([#327])
-
+* reorganize OSM coordinate constants ([#320])
```
tyrasd, line 25:
```suggestion
* reorganize OSM coordinate conversion factors. ([#320])

```

oshdb-tool/etl/src/main/java/org/heigit/ohsome/oshdb/tool/importer/extract/collector/StatsCollector.java
```diff
@@ -123,7 +124,7 @@ public void print(PrintStream out){
     }
     
     if(minLon != Integer.MAX_VALUE){
-      out.printf("data.bbox=%.7f,%.7f,%.7f,%.7f%n",minLon*OSHDB.GEOM_PRECISION,minLat*OSHDB.GEOM_PRECISION,maxLon*OSHDB.GEOM_PRECISION,maxLat*OSHDB.GEOM_PRECISION);
+      out.printf("data.bbox=%.7f,%.7f,%.7f,%.7f%n",minLon*GEOM_PRECISION,minLat*GEOM_PRECISION,maxLon*GEOM_PRECISION,maxLat*GEOM_PRECISION);
```
tyrasd, line 127: code style:

```suggestion
      out.printf("data.bbox=%.7f,%.7f,%.7f,%.7f%n", minLon * GEOM_PRECISION,
          minLat * GEOM_PRECISION, maxLon * GEOM_PRECISION, maxLat * GEOM_PRECISION);
```

oshdb-tool/etl/src/main/java/org/heigit/ohsome/oshdb/tool/importer/util/ZGrid.java
```diff
@@ -165,19 +167,19 @@ private static long morton(long x, long y) {
   }
 
   private static long normalizeLon(long lon) {
-    return lon + (long) (180.0 * OSHDB.GEOM_PRECISION_TO_LONG);
+    return lon + (long) (180.0 * GEOM_PRECISION_TO_LONG);
```
tyrasd, line 170: why not 
```suggestion
    return lon + coordinateToLong(180.0);
```
?

oshdb/src/main/java/org/heigit/ohsome/oshdb/util/OSHDBBoundingBox.java
```diff
@@ -1,17 +1,16 @@
...
 
 public class OSHDBBoundingBox implements Serializable {
   private static final long serialVersionUID = 1L;
-  public static final OSHDBBoundingBox EMPTY = new OSHDBBoundingBox(0L, 0L, 0L, 0L);
-  public static final OSHDBBoundingBox INVALID = new OSHDBBoundingBox(Long.MAX_VALUE,Long.MAX_VALUE,Long.MIN_VALUE,Long.MIN_VALUE);
-
 
   /**
-   * calculates the intersection of two bounding boxes
+   * calculates the intersection of two bounding boxes.
```
tyrasd, line 13:
```suggestion
   * Calculates the intersection of two bounding boxes.
```

tyrasd (commented):
I noticed that sometimes we perform `Math.round` when converting from double to long, and some other times we only cast to `(double)`. Should we be more consistent or does it not matter?

tyrasd:
* [ ] todo: https://github.com/GIScience/oshdb/pull/369/files#r618290231
  > So my suggestion here is, to remove the double return value methods completely and rename/change the long return value variants to
  > 
  > ```
  > int minLon();
  > int minLat();
  > int maxLon();
  > int maxLat();
  > ```

PR #322, opened 2020-12-18
make ohsome filters work with groupByEntity queries

### LINKED ISSUE #321: ohsome filters don't work with groupByEntity queries
**Describe the bug**
`filter(String)` does not work together with `groupByEntity`

**To Reproduce**

```java
Number osmTypeFilterResult = mapreducer.groupByEntity().osmType(OSMType.WAY).count();
Number stringFilterResult = mapreducer.groupByEntity().filter("type:way").count();

assertEquals(osmTypeFilterResult, stringFilterResult);
```

**Expected behavior**
The two queries above should both produce the same result.

**Error messages, logs, screenshots**
```
java.lang.ClassCastException
Caused by: java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class org.heigit.ohsome.oshdb.api.object.OSMEntitySnapshot
```

**Affected versions:**
 - OSHDB Version 0.6.0
 - OSHDB Version 0.7.0-SNAPSHOT (current master branch)

tyrasd:
For now this just contains a failing unit test to showcase the problem. After this is resolved, we should create patch release `0.6.1` together with #326 (which needs to be backported in `0.5.11` as well).

## Type of change

## Corresponding issue
Closes #321

# Checklist
- ~~I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) if necessary~~

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
+++ b/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
@@ -1,7 +1,6 @@
 package org.heigit.ohsome.oshdb.api.mapreducer;
 
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
 import com.tdunning.math.stats.TDigest;
@@ -759,16 +758,19 @@ public abstract class MapReducer<X> implements
       if (ret.forClass.equals(OSMContribution.class)) {
         @SuppressWarnings("unchecked") MapReducer<X> result = (MapReducer<X>)
             ret.map(x -> (Collection<OSMContribution>) x)
-                .map(contributions -> Lists.newArrayList(Iterables.filter(contributions, c ->
-                    c != null && (f.applyOSMGeometry(c.getEntityBefore(), c::getGeometryBefore)
-                    || f.applyOSMGeometry(c.getEntityAfter(), c::getGeometryAfter)))))
+                .map(contributions -> contributions.stream()
+                    .filter(c -> c != null
+                        && (f.applyOSMGeometry(c.getEntityBefore(), c::getGeometryBefore)
+                        || f.applyOSMGeometry(c.getEntityAfter(), c::getGeometryAfter)))
+                    .collect(Collectors.toCollection(ArrayList::new)))
                 .filter(contributions -> !contributions.isEmpty());
         ret = result;
       } else if (ret.forClass.equals(OSMEntitySnapshot.class)) {
         @SuppressWarnings("unchecked") MapReducer<X> result = (MapReducer<X>)
             ret.map(x -> (Collection<OSMEntitySnapshot>) x)
-                .map(snapshots -> Lists.newArrayList(Iterables.filter(snapshots, s ->
-                    s != null && f.applyOSMGeometry(s.getEntity(), s::getGeometry))))
+                .map(snapshots -> snapshots.stream().filter(s ->
+                    s != null && f.applyOSMGeometry(s.getEntity(), s::getGeometry))
+                    .collect(Collectors.toCollection(ArrayList::new)))
                 .filter(snapshots -> !snapshots.isEmpty());
         ret = result;
       }
--- a/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
+++ b/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
@@ -1,7 +1,7 @@
 package org.heigit.ohsome.oshdb.api.mapreducer;
 
-import com.google.common.collect.Collections2;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
 import com.tdunning.math.stats.TDigest;
@@ -759,7 +759,7 @@ public abstract class MapReducer<X> implements
       if (ret.forClass.equals(OSMContribution.class)) {
         @SuppressWarnings("unchecked") MapReducer<X> result = (MapReducer<X>)
             ret.map(x -> (Collection<OSMContribution>) x)
-                .map(contributions -> new ArrayList<>(Collections2.filter(contributions, c ->
+                .map(contributions -> Lists.newArrayList(Iterables.filter(contributions, c ->
                     c != null && (f.applyOSMGeometry(c.getEntityBefore(), c::getGeometryBefore)
                     || f.applyOSMGeometry(c.getEntityAfter(), c::getGeometryAfter)))))
                 .filter(contributions -> !contributions.isEmpty());
@@ -767,7 +767,7 @@ public abstract class MapReducer<X> implements
       } else if (ret.forClass.equals(OSMEntitySnapshot.class)) {
         @SuppressWarnings("unchecked") MapReducer<X> result = (MapReducer<X>)
             ret.map(x -> (Collection<OSMEntitySnapshot>) x)
-                .map(snapshots -> new ArrayList<>(Collections2.filter(snapshots, s ->
+                .map(snapshots -> Lists.newArrayList(Iterables.filter(snapshots, s ->
                     s != null && f.applyOSMGeometry(s.getEntity(), s::getGeometry))))
                 .filter(snapshots -> !snapshots.isEmpty());
         ret = result;
--- a/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
+++ b/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
@@ -754,9 +754,32 @@ public abstract class MapReducer<X> implements
     // apply geometry filter as first map function
     final List<MapFunction> remainingMappers = List.copyOf(ret.mappers);
     ret.mappers.clear();
-    if (this.grouping == Grouping.BY_ID) {
-      if (ret.forClass.equals(OSMContribution.class)) {
-        @SuppressWarnings("unchecked") MapReducer<X> result = (MapReducer<X>)
+    if (this.grouping == Grouping.NONE) {
+      // no grouping -> directly filter using the geometries of the snapshot / contribution
+      if (ret.forClass.equals(OSMEntitySnapshot.class)) {
+        ret = ret.filter(x -> {
+          OSMEntitySnapshot s = (OSMEntitySnapshot) x;
+          return f.applyOSMGeometry(s.getEntity(), s::getGeometry);
+        });
+      } else if (ret.forClass.equals(OSMContribution.class)) {
+        ret = ret.filter(x -> {
+          OSMContribution c = (OSMContribution) x;
+          return f.applyOSMGeometry(c.getEntityBefore(), c::getGeometryBefore)
+              || f.applyOSMGeometry(c.getEntityAfter(), c::getGeometryAfter);
+        });
+      }
+    } else if (this.grouping == Grouping.BY_ID) {
+      // grouping by entity -> filter each list entry individually
+      if (ret.forClass.equals(OSMEntitySnapshot.class)) {
+        @SuppressWarnings("unchecked") MapReducer<X> filteredListMapper = (MapReducer<X>)
+            ret.map(x -> (Collection<OSMEntitySnapshot>) x)
+                .map(snapshots -> snapshots.stream().filter(s ->
+                    s != null && f.applyOSMGeometry(s.getEntity(), s::getGeometry))
+                    .collect(Collectors.toCollection(ArrayList::new)))
+                .filter(snapshots -> !snapshots.isEmpty());
+        ret = filteredListMapper;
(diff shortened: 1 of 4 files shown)


oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -754,20 +755,41 @@ protected MapReducer(MapReducer<?> obj) {
...
-    } else if (ret.forClass.equals(OSMEntitySnapshot.class)) {
-      ret = ret.filter(x -> {
-        OSMEntitySnapshot s = (OSMEntitySnapshot) x;
-        return f.applyOSMGeometry(s.getEntity(), s::getGeometry);
-      });
+    if (this.grouping == Grouping.BY_ID) {
+      if (ret.forClass.equals(OSMContribution.class)) {
+        @SuppressWarnings("unchecked") MapReducer<X> result = (MapReducer<X>)
+            ret.map(x -> (Collection<OSMContribution>) x)
+                .map(contributions -> Collections2.filter(contributions, c -> c != null
```
tyrasd, line 762: todo: check if it's save to assume that this is still a `List` (not just a generic `Collection` which doesn't implement stuff like `get(index)`

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOhsomeFilter.java
```diff
@@ -57,8 +57,7 @@ public void testFilterString() throws Exception {
 
   @Test
   public void testFilterObject() throws Exception {
-    MapReducer<OSMEntitySnapshot> mr = createMapReducer();
-    Number result = mr
+    Number result = createMapReducer()
```
joker234, line 60: Is this a cleanup that changes or affects anything apart from syntax cleanup?

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOhsomeFilter.java
```diff
@@ -79,12 +78,20 @@ public void testAggregateFilter() throws Exception {
...
         .filter(filterParser.parse("(geometry:polygon or geometry:other) and building=*"))
         .count();
 
     assertEquals(42, result.get(OSMType.WAY).intValue());
   }
+
+  @Test
+  public void testFilterGroupByEntity() throws Exception {
+    MapReducer<OSMEntitySnapshot> mr = createMapReducer();
+    Number osmTypeFilterResult = mr.groupByEntity().osmType(OSMType.WAY).count();
```
joker234, line 92: `osmType(…)` is deprecated. What would be the plan for this test, if the method is removed in the future? Maybe add a comment to this test explaining what needs to be done if this method is removed.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -754,17 +754,44 @@ protected MapReducer(MapReducer<?> obj) {
...
-        return f.applyOSMGeometry(s.getEntity(), s::getGeometry);
-      });
+    if (this.grouping == Grouping.NONE) {
+      // no grouping -> directly filter using the geometries of the snapshot / contribution
+      if (ret.forClass.equals(OSMEntitySnapshot.class)) {
+        ret = ret.filter(x -> {
+          OSMEntitySnapshot s = (OSMEntitySnapshot) x;
+          return f.applyOSMGeometry(s.getEntity(), s::getGeometry);
+        });
+      } else if (ret.forClass.equals(OSMContribution.class)) {
```
joker234, line 764: not covered by any tests

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -754,17 +754,44 @@ protected MapReducer(MapReducer<?> obj) {
...
+      // grouping by entity -> filter each list entry individually
+      if (ret.forClass.equals(OSMEntitySnapshot.class)) {
+        @SuppressWarnings("unchecked") MapReducer<X> filteredListMapper = (MapReducer<X>)
+            ret.map(x -> (Collection<OSMEntitySnapshot>) x)
+                .map(snapshots -> snapshots.stream().filter(s ->
+                    s != null && f.applyOSMGeometry(s.getEntity(), s::getGeometry))
+                    .collect(Collectors.toCollection(ArrayList::new)))
+                .filter(snapshots -> !snapshots.isEmpty());
+        ret = filteredListMapper;
+      } else if (ret.forClass.equals(OSMContribution.class)) {
```
joker234, line 781: not covered by any tests

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -754,17 +754,44 @@ protected MapReducer(MapReducer<?> obj) {
...
+                .map(contributions -> contributions.stream()
+                    .filter(c -> c != null
+                        && (f.applyOSMGeometry(c.getEntityBefore(), c::getGeometryBefore)
+                        || f.applyOSMGeometry(c.getEntityAfter(), c::getGeometryAfter)))
+                    .collect(Collectors.toCollection(ArrayList::new)))
+                .filter(contributions -> !contributions.isEmpty());
+        ret = filteredListMapper;
+      }
+    } else {
+      throw new UnsupportedOperationException(
```
joker234, line 793: Is this an Exception that can actually happen? I think no, so I suppose you cannot test it. But we should keep it, if someone extends the `Grouping` enum without adding a case here.

joker234 (commented):
First remarks. I'll have a look in the new contents of `else if (this.grouping == Grouping.BY_ID)` later.

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOhsomeFilter.java
tyrasd: just a minor cleanup, when the test only needs one mapreducer, we don't need the local variable. :shrug:

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOhsomeFilter.java
tyrasd: right. that would be one of the todos in #324

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
tyrasd: huh, true. Strictly speaking this part of the code wasn't really changed in this PR. but I guess it would be good to have some basic test for this path as well.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
tyrasd: no, it can't happen at the moment because the enum only has two possibilities. At first, I had implemented this with a `switch` (with `default` to be the same as `case NONE`), but then sonarlint complained that one shouldn't use switch statements with only two branches :shrug: I thought by adding the "hypothetical" throw statement here, it's clear that something must be added if at some point there would be an additional _grouping_ possibility.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
tyrasd: oh gosh, this branch was actually pretty buggy and slipped through because it wasn't covered by the tests. :see_no_evil: thanks for spotting that!

should be fixed in d468552e

joker234:
Additional remark: Should we rename `TestOhsomeFilter.java` to `TestOSHDBFilter.java`?

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
tyrasd: fixed in d468552

tyrasd:
> Additional remark: Should we rename `TestOhsomeFilter.java` to `TestOSHDBFilter.java`?

sure, that wouldn't hurt. :arrow_right: 170cc32f

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
joker234: So let's keep that in mind for future versions of the oshdb :+1:

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -754,17 +755,57 @@ protected MapReducer(MapReducer<?> obj) {
...
+      } else if (ret.forClass.equals(OSMContribution.class)) {
+        ret = ret.filter(x -> {
+          OSMContribution c = (OSMContribution) x;
+          if (c.is(ContributionType.CREATION)) {
+            return f.applyOSMGeometry(c.getEntityAfter(), c::getGeometryAfter);
+          } else if (c.is(ContributionType.DELETION)) {
+            return f.applyOSMGeometry(c.getEntityBefore(), c::getGeometryBefore);
+          } else {
+            return f.applyOSMGeometry(c.getEntityBefore(), c::getGeometryBefore)
+                || f.applyOSMGeometry(c.getEntityAfter(), c::getGeometryAfter);
```
joker234, line 774: According to my local tests, the lines 770 to 774 are still not tested. They look sane to me, so I wouldn't require it, but maybe it makes sense to do that anyway.

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOSHDBFilter.java
```diff
@@ -33,21 +35,34 @@
...
+  private MapReducer<OSMEntitySnapshot> createMapReducerOSMEntitySnapshot() throws Exception {
     return OSMEntitySnapshotView.on(oshdb)
         .areaOfInterest(bbox)
         .timestamps("2014-01-01");
   }
 
+  private MapReducer<OSMContribution> createMapReducerOSMContribution() throws Exception {
+    return OSMContributionView.on(oshdb)
+        .areaOfInterest(bbox)
+        .timestamps("2008-01-01", "2014-01-01");
```
joker234, line 53: With this time range, I suppose the idea is that `createMapReducerOSMContribution()` contains the same result as `createMapReducerOSMEntitySnapshot()`. I think that's okay for now since the contribution endpoint is tested properly in other oshdb-api tests, correct?

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOSHDBFilter.java
```diff
@@ -79,12 +93,26 @@ public void testAggregateFilter() throws Exception {
...
+    SortedMap<OSMType, Integer> result = createMapReducerOSMEntitySnapshot()
         .aggregateBy(x -> x.getEntity().getType())
         .filter(filterParser.parse("(geometry:polygon or geometry:other) and building=*"))
         .count();
 
     assertEquals(42, result.get(OSMType.WAY).intValue());
   }
+
+  @Test
+  public void testFilterGroupByEntity() throws Exception {
```
joker234, line 105: Could be split in two separate tests. If it would contain more than the two asserts, SonarLint would recommend that, but I think it's okay for now.

joker234 (approved):
Some minor comments and suggestions. None of them needs to be addressed necessarily to merge this PR.

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOSHDBFilter.java
tyrasd: yes, that was the idea. since there are no matching-but-later-deleted elements in this test data set, the results should match.

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOSHDBFilter.java
tyrasd: sure, but I think this is good enough for now ;)

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
tyrasd: if we had a properly crafted test data set, it would be easy to trigger these branches as well. That would be nice to have eventually. But I think we can live with the current implementation for now, right?

PR #331, opened 2021-01-05
link to mentionend issues or PRs

tyrasd:


joker234 (approved):
Is it possible to collect the links at the end of the file instead of for each version separately?

tyrasd:
> Is it possible to collect the links at the end of the file instead of for each version separately?

I think it's better to have them locally at the end of the section of each version because that way one doesn't have to modify stuff at the very top and at the very bottom of the file when adding something to the changelog.

joker234:
> Possible: Yes, but I think it's better to have them locally at the end of the section of each version because that way one doesn't have to modify stuff at the very top and at the very bottom of the file when adding something to the changelog.

But sometimes you'll have the case that you want to reuse links, like here: https://github.com/GIScience/oshdb/pull/331/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR86

But this was just a comment, you can merge the PR, if you want.

PR #332, opened 2021-01-11
clean up OSHEntities utility class

### LINKED ISSUE #327: clean OSHEntities class (get*Timestamps methods)
These methods have not aged very well, yet are a crucial part of the whole OSHDB processing pipeline. They should be be better documented (inline comments, javadoc), deserve better test [coverage](https://jenkins.ohsome.org/job/oshdb/job/master/380/jacoco/org.heigit.ohsome.oshdb.osh/OSHEntities/), code style issues should be resolved, and it would be nice if they were a bit more organized (maybe we should even move to oshdb-utils module).

Related issues:
* #325 (PR #326)
* PR #10
* PR #12

tyrasd:
improved  various stuff about the OSHEntities utility method class:

* [x] move `getModificationTimestamps` and `getChangesetTimestamps` methods to oshdb-utils module
* [x] add missing javadoc
* [x] fix checkstyle issues
* [x] add inline comments
* [x] improve test coverage

### Corresponding issue
Closes #327

### Checklist
- [x] I have commented my code :smiley: 


oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
```diff
@@ -802,6 +804,22 @@ public IterateAllEntry(
...
+    Iterator<? extends OSMEntity> itr = osh.getVersions().iterator();
+    while (itr.hasNext() && i >= 0) {
+      OSMEntity osm = itr.next();
+      if (osm.getTimestamp().getRawUnixTimestamp() <= byTimestamps.get(i).getRawUnixTimestamp()) {
+        while (i >= 0 && osm.getTimestamp().getRawUnixTimestamp() <= byTimestamps.get(i)
+            .getRawUnixTimestamp()) {
+          result.put(byTimestamps.get(i), osm);
+          i--;
+        }
+      }
```
rtroilo, line 821: I think the first if check is unnecessary, we can reduce the nesting by one level. And for the timestamp comparison, why not use the compareTo method of the OSHDBTimestamp directly.
Here my suggestion:
```java
while (itr.hasNext() && i >= 0) {
  OSMEntity osm = itr.next();
  while (i >= 0 && osm.getTimestamp().compareTo(byTimestamps.get(i)) <= 0) {
    result.put(byTimestamps.get(i), osm);
    i--;
  }
}
```

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,357 @@
...
+  private static Map<OSHDBTimestamp, Long> getChangesetTimestamps(OSHNode osh) {
+    Map<OSHDBTimestamp, Long> result = new TreeMap<>();
+    osh.getVersions().forEach(osm -> result.put(osm.getTimestamp(), osm.getChangesetId()));
+    return result;
+  }
+
+  private static Map<OSHDBTimestamp, Long> getChangesetTimestamps(OSHWay osh) {
+    Map<OSHDBTimestamp, Long> result = new TreeMap<>();
+
+    List<OSMWay> ways = org.heigit.ohsome.oshdb.osh.OSHEntities.toList(osh.getVersions());
```
rtroilo, line 66: Why do you use the full-qualified name for OSHEntities, instead of an import?

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,357 @@
...
+      default:
+        throw new UnsupportedOperationException(format(UNSUPPORTED_OSMTYPE_MESSAGE, osh.getType()));
+    }
+  }
+
+  private static List<OSHDBTimestamp> getModificationTimestamps(OSHNode osh) {
+    List<OSHDBTimestamp> result = new ArrayList<>();
+    for (OSMEntity osm : osh.getVersions()) {
+      result.add(osm.getTimestamp());
+    }
```
rtroilo, line 144: we should size the ArrayList to avoid growing. like it is done in `OSHEntities.toList`

```java
var itr = osh.getVersions().iterator();
if (!itr.hasNext()) {
  return Collections.emptyList();
}
var osm = itr.next();
var result = new ArrayList<OSHDBTimestamp>(osm.getVersion());
result.add(osm.getTimestamp());
while (itr.hasNext()) {
  osm = itr.next();
  result.add(osm.getTimestamp());
}
```
```

File: oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java [hunk 4]
```diff
@@ -802,6 +804,22 @@ public IterateAllEntry(
...
 
-}
-
-
+  private static SortedMap<OSHDBTimestamp, OSMEntity> getVersionsByTimestamps(
+      OSHEntity osh, List<OSHDBTimestamp> byTimestamps) {
+    SortedMap<OSHDBTimestamp, OSMEntity> result = new TreeMap<>();
+
+    int i = byTimestamps.size() - 1;
+    Iterator<? extends OSMEntity> itr = osh.getVersions().iterator();
```
rtroilo, line 812: we should check if we need to have a map returned instead of an sorted list. The versions from `osh.getVersions()` are already sorted by version_number in decreasing, maybe we can use this fact.
I checked the call sites for getVersionsbyTimestamp and in the current usage, it will only be iterated over the map.
So returning here a list instead of a map could be better performance wise.

In the moment, because `osh.getVersions()` returns timestamps in  decreasing, but the TreeMap sorts the timestamps in increasing order, I think, there are far more checks then using a list, and sort afterwards (and remove duplicates )

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,357 @@
...
+import org.heigit.ohsome.oshdb.osm.OSMEntity;
+import org.heigit.ohsome.oshdb.osm.OSMMember;
+import org.heigit.ohsome.oshdb.osm.OSMNode;
+import org.heigit.ohsome.oshdb.osm.OSMRelation;
+import org.heigit.ohsome.oshdb.osm.OSMWay;
+import org.heigit.ohsome.oshdb.util.OSHDBTimestamp;
+
+public class OSHEntityTimeUtils {
+  private OSHEntityTimeUtils() {
+    throw new IllegalStateException("utility class");
```
rtroilo, line 33: any reasons to throw an exception in a private constructor? No one can call it any where accept the class it self.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,357 @@
...
+  public static Map<OSHDBTimestamp, Long> getChangesetTimestamps(OSHEntity osh) {
+    switch (osh.getType()) {
+      case NODE:
+        return getChangesetTimestamps((OSHNode) osh);
+      case WAY:
+        return getChangesetTimestamps((OSHWay) osh);
+      case RELATION:
+        return getChangesetTimestamps((OSHRelation) osh);
+      default:
+        throw new UnsupportedOperationException(format(UNSUPPORTED_OSMTYPE_MESSAGE, osh.getType()));
```
rtroilo, line 53: I also often see in such a case an IlegalStateException as well. I think we should just check what is common to throw in such a case.

- do we not support an unknown osmtype
- does the osh have an illegal state because of the unknow osmtype

both are valid views i guess

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,357 @@
...
+    Map<OSHDBTimestamp, Long> result = new TreeMap<>();
+
+    List<OSMRelation> rels = org.heigit.ohsome.oshdb.osh.OSHEntities.toList(osh.getVersions());
+    rels.forEach(osmRel -> result.put(osmRel.getTimestamp(), osmRel.getChangesetId()));
+
+    // recurse rel members
+    try {
+      Stream.concat(osh.getNodes().stream(), osh.getWays().stream())
+          .filter(Objects::nonNull)
+          .forEach(oshEntity -> getChangesetTimestamps(oshEntity).forEach(result::putIfAbsent));
```
rtroilo, line 96: in `getChangesetTimestamps(OSHWay osh)` we use a for-each loop instead of stream for the members.
To make this more consistent, me suggestion would be:
```java
for (var mem : osh.getNodes()) {
  if (mem == null) {
    continue;
  }
  for (var version : mem.getVersions()) {
    result.putIfAbsent(version.getTimestamp(), version.getChangesetId());
  }
}

for (var mem : osh.getWays()) {
  if (mem == null) {
    continue;
  }
  getChangesetTimestamps(mem).forEach(result::putIfAbsent);
}
```

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,357 @@
...
+  public static List<OSHDBTimestamp> getModificationTimestamps(OSHEntity osh,
+      Predicate<OSMEntity> osmEntityFilter, Map<OSHDBTimestamp, Long> changesetTimestamps) {
+    List<OSHDBTimestamp> allModificationTimestamps = getModificationTimestamps(osh, true, osmEntityFilter);
+
+    if (allModificationTimestamps.size() <= 1) {
+      return allModificationTimestamps;
+    }
+    // group modification timestamps by changeset
+    List<OSHDBTimestamp> result = new ArrayList<>(allModificationTimestamps.size());
+    allModificationTimestamps = Lists.reverse(allModificationTimestamps);
```
rtroilo, line 175: `getModifcationTimestamps(..)` reversed the list of timestamp, and here we reverse it back, maybe we should not reverse in the first place or let's pass parameter in which order we need the modificationTimestamps, so that there is no back and forth.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: This is suggested by sonarlint ([rule S1118](https://sonarcloud.io/organizations/default/rules?languages=java&open=java%3AS1118&q=S1118)).  I believe the reasoning behind it might be that under some strange conditions this constructor might still be called (idk, maybe through _reflection_ or something similar).

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: yeah, I thought about that as well. I found [this thread](https://stackoverflow.com/a/49749857/1627467) on SO:

> `UnsupportedOperationException` should be used as the method is not supported at all while `IllegalStateException` should be used as the method is supported but that in the current state, it is not legal.

They give the example of the `Iterator.remove` method which can be "unsupported" for some kinds of iterators (throwing `UnsupportedOperationException`), but might also throw `IllegalStateException` when they run into a situation where the method is implemented, but it's not possible to remove any more items because there are no more to remove.

In the end it doesn't really matter since this line cannot be triggered by "normal" code anyway since there are only the above listed 3 `OSMType`s, and the default case is just added to make the Java (11) compiler happy. If we eventually upgrade our Java version (to ⪆ 14), we could get rid of the `default` clause here (see [JEP 361](https://openjdk.java.net/jeps/361#Exhaustiveness).

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,357 @@
...
+    long nextChangeset = -1L;
+    for (OSHDBTimestamp timestamp : allModificationTimestamps) {
+      long changeset = changesetTimestamps.get(timestamp);
+      if (changeset != nextChangeset) {
+        result.add(timestamp);
+      }
+      nextChangeset = changeset;
+    }
+
+    return Lists.reverse(result);
```
rtroilo, line 185: We could try to inplace group modification timestamps, to avoid the additional array creation and the reversing back and forth.
Here an untested suggestion:
```java
int pos = 0;
for (int i=0; i < allModificationTimestamps.size() ;) {
  var timestamp = allModificationTimestamps.get(i);
  var changeset = changesetTimestamps.get(timestamp);
  i++;
  for (; i < allModificationTimestamps.size(); i++) {
    var nextTimestamp = allModificationTimestamps.get(i);
    var nextChangeset = changesetTimestamps.get(nextTimestamp);
    if (changeset != nextChangeset) {
      break;
    }
    timestamp = nextTimestamp;
  }
  allModificationTimestamps.set(pos++, timestamp);
}
return allModificationTimestamps.subList(0, pos);
```

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: Huh. It might be an artifact of the semi-automated refactoring done by my IDE when moving the [method](https://github.com/GIScience/oshdb/pull/332/files#diff-ee965dbc15238b368d4cbd6a2377ac995b40270c97a077fb5b64d21fff9d602aL134) to this new class. Thanks for catching this, I'll change it to an import. :+1:

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: I didn't like the duplication too much when done exactly this way.

Instead, I will replace it with a new helper method which can be shared between `getChangesetTimestamps(OSHWay)` and `getChangesetTimestamps(OSHRelation)`.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: good suggestion. I made `OSHEntities.toList` a bit more powerful to accommodate  for this use case as well.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,360 @@
...
+    if (!recurse || osh instanceof OSHNode) {
+      return Lists.reverse(entityTimestamps);
+    }
+    // recurse members: start by collecting all referenced members' validity time intervals
+    var membersValidityTimes = collectMembershipTimeIntervals(osh, osmEntityFilter);
+    // fill in member modification timestamps
+    var result = fillMembersModificationTimestamps(membersValidityTimes);
+    // combine with the entity's own modification timestamps
+    result.addAll(entityTimestamps);
+    return new ArrayList<>(result);
```
rtroilo, line 194: what caught my eye here, is that in the non-recurse or node case we return a reversed list. But not in the other case. I wonder if this is correct.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: ah, I forgot/overlooked that the second loop need to call `getChangesetTimestamps`. :see_no_evil:

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
tyrasd: Good suggestion!

A quick test revealed that a simple `Lists.reverse()` is sufficient (at least to make the unit tests pass), so perhaps one doesn't even need to actually sort and/or deduplicate the result.

But I would be slightly hesitant to include this change in this PR, since the goal was not to do any major changes or performance improvements. It was just to make the code better organized, cleaned up and better documented. I would however appreciate it if you could create a separate issue for this suggestion, than we can work on it after this PR is merged.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
tyrasd: :thinking: We have to check this thoroughly. At least the unit tests manage to hit both cases of the `if` statement:

image

tyrasd: ah, wait. I see what you mean now :see_no_evil:  The `while` condition is equivalent to the surrounding `if`. Sure, then it can certainly be omitted.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: That _should_ not be the case, since:

* in the `!recurse || osh instanceof OSHNode` case the result is `Lists.reverse` of the result from `getModificationTimestampsReverseNonRecursed` which is in reverse order :arrow_right: we return it in the "natural" order.
* otherwise the `result` is a list generated from a `SortedSet` (result) which takes care of sorting for us. :arrow_right: same "natural" order as above

Maybe it would be better to not use `var result = …`  here, so it's clearer that the result stems from a sorted set? What do think?

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
rtroilo: Good point, I will then open a new issue/pr for improving the way we generate the contribution. 👍

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
rtroilo: Ah, I see. The rule is about specify a private constructor, so that the default public is not generated, but the exception within is not necessary but also not harmful.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
tyrasd: Merci. Btw: feel free to open an issue already now, then you don't have to keep it in mind for too long. :wink: 

Also, I have already an almost complete implementation of your suggestion in my `git stash`, in case you want to skip the boring implementation details.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: good idea, but what do you think about this (IMHO) slightly less complex solution:

```java

    // group modification timestamps by changeset
    long nextChangeset = -1L;
    int pos = allModificationTimestamps.size();
    for (OSHDBTimestamp timestamp : Lists.reverse(allModificationTimestamps)) {
      long changeset = changesetTimestamps.get(timestamp);
      if (changeset != nextChangeset) {
        allModificationTimestamps.set(--pos, timestamp);
      }
      nextChangeset = changeset;
    }

    return allModificationTimestamps.subList(pos, allModificationTimestamps.size());
```

it comes at the cost of one additional `Lists.reverse`, but since that's just a lightweight view, I guess we ca live with it.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
rtroilo: could we omitted that redundant if check, and use compareTo from the OSHDBTimestamp instead checking against the rawUnixTimestamp?

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  /**
+   * Returns for each member of an entity a list of time intervals when the belong to the entity
+   * matching the given predicate.
+   */
+  private static Map<OSHEntity, LinkedList<OSHDBTimestamp>> collectMembershipTimeIntervals(
+      OSHEntity osh, Predicate<OSMEntity> osmEntityFilter) {
+    Map<OSHEntity, LinkedList<OSHDBTimestamp>> memberTimes
+        = new TreeMap<>(Comparator.comparingLong(OSHEntity::getId));
```
rtroilo, line 267: A TreeMap is not requiered at this point, so a HashMap would fit better here.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
tyrasd: yes, see https://github.com/GIScience/oshdb/pull/332/commits/5a26b21633d3df7785e98b7aa1e1737bf3fcc5c1#diff-fbb13d5279408199848215dfdc8b65baafc3725c8a64c579d9ca119aeba12cc6R819

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/osh/OSHEntityTimeUtils.java
tyrasd: Am I right that the benefit of `HashMap` (over other `Map` implementations) would be faster access when calling `.get` below?

rtroilo: That's true. for a tree map you have to do potential more checks than with a HashMap.
We should also keep in mind, that it could be worth, to check if the members actually change between one major-version to the next, I'm not sure how much of an impact this has. Could be something for the next PR.

tyrasd: PS: Could it be an issue for using a HashMap that OSHEntity does not implement hashCode and equals?

>  We should also keep in mind, that it could be worth, to check if the members actually change between one major-version to the next, I'm not sure how much of an impact this has. Could be something for the next PR.

I'm not 100% sure if I understand what you propose, but this method does indeed not look at what happens to the members, it just collects the time intervals where they are a member of the entity at hand. The actual _change_s (modificationTimestamps) are calculated in `fillMembersModificationTimestamps` below.

rtroilo: > Could it be an issue for using a HashMap that OSHEntity does not implement hashCode and equals?
Yes, very good point, that's could be an issue, normally! 
Every object inherit the hashcode, equals from the Object class, which do a check an the "objects address". 
But in this case the returned  osh-entity from member.getEntity() is always the same object.

rtroilo: Actually, good that we think about this
```java
Map<OSHEntity, LinkedList<OSHDBTimestamp>> memberTimes
        = new TreeMap<>(Comparator.comparingLong(OSHEntity::getId));
```

is a potential bug in case the osh argument of ` collectMembershipTimeIntervals` is of type relation!
It is possible that ( and I know that here are cases), that a relation has a way member with the same id with a node member.

tyrasd: you are right! :scream: good catch! I will switch to the HashMap as suggested.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
rtroilo: Cool, thank you looking forward to work together with you!

rtroilo (approved):
Thank you @tyrasd , great work and was nice experience with you.

PR #263, opened 2020-08-28
specify port range for the testing ignite instance

tyrasd:
This should allow unit tests to be run in parallel without interfering with each other. (hopefully :crossed_fingers: )

PR #335, opened 2021-01-20
don't manipulate cluster state

tyrasd:
The OSHDB-API should not change the state of an already running ignite cluster. If it is inactive, it might be for a reason, if it is in read-write mode, it might also be in that state for a reason, etc.

This is a regression in OSHDB version 0.6.0 introduced with #263.

### Checklist
- ~~I have commented my code~~

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/db/OSHDBIgnite.java
+++ b/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/db/OSHDBIgnite.java
@@ -8,7 +8,6 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.ignite.Ignite;
 import org.apache.ignite.Ignition;
-import org.apache.ignite.cluster.ClusterState;
 import org.apache.ignite.lang.IgniteRunnable;
 import org.heigit.ohsome.oshdb.api.mapreducer.MapReducer;
 import org.heigit.ohsome.oshdb.api.mapreducer.backend.MapReducerIgniteAffinityCall;
@@ -40,7 +39,6 @@ public class OSHDBIgnite extends OSHDBDatabase implements AutoCloseable {
 
   public OSHDBIgnite(Ignite ignite) {
     this.ignite = ignite;
-    this.ignite.cluster().state(ClusterState.ACTIVE_READ_ONLY);
   }
 
   public OSHDBIgnite(String igniteConfigFilePath) {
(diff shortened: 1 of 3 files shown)


oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/db/OSHDBIgnite.java
```diff
@@ -40,7 +39,6 @@ public OSHDBIgnite() {
 
   public OSHDBIgnite(Ignite ignite) {
     this.ignite = ignite;
-    this.ignite.cluster().state(ClusterState.ACTIVE_READ_ONLY);
```
rtroilo, line 43: We should actually check the current cluster().state() and throw an exception right here if it is not in active or active_read_only state.
Because in a case when the cluster is actually not active, there will be probably an exception on the first cache operation somewhere which would then be unexpected. 
I think it's better to notify as soon as possible.

tyrasd: can you please suggest an implementation for this?

tyrasd: PS: what if the cluster goes active/inactive between the time where the connection gets established and when the actual OSHDB query is performed?

rtroilo: > can you please suggest an implementation for this?

sure here are 2 suggestions:

```java
  // explicit check for good states
  if (!EnumSet.of(ACTIVE, ACTIVE_READ_ONLY).contains(cluster.state())) {
    // throw exception about cluster is not ready (yet)! 
  }
  // or check for the fail state
  if (cluster.state() == INACTIVE) {
    // throw exception about cluster is not ready (yet)! 
  }
```
I think the first suggestion is more robust in case there will be in the future more and other ClusterStates

rtroilo: > PS: what if the cluster goes active/inactive between the time where the connection gets established and when the actual OSHDB query is performed?

When it goes inactive while a OSHDB query is performed, probably on the next cache operation there could be an exception similar to 
> class org.apache.ignite.internal.processors.cache.distributed.dht.IgniteClusterReadOnlyException: Failed to finish operation. Cluster in read-only mode!

We can check what specific exception will be thrown if any at all. I will try it in the test cluster. And report it back here

rtroilo (approved):

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/db/OSHDBIgnite.java
joker234: For reference: will be implemented in #337

PR #340, opened 2021-01-22
update DOI for version 0.6.2

joker234:
### Description
updated DOI in README for version 0.6.2


joker234:
we have the general DOI in Readme

PR #343, opened 2021-01-27
Fix incomplete touching inners

tyrasd:
### Description

Fixes a bug which caused an infinite loop in the `cutAtSement` method (used in `mergeTouchingRings` routine) because it is called with non-closed linestrings instead of rings. Fixed by making sure that `buildRings` does only return valid rings.

This has the side effect that some errorneous multipolygons now get "cleaned" up slightly into (smaller) polygons, where they were returned as `GeometryCollections` before (prior to `0.6.0`). But since the old (`buffer(0)`, see #249) implementation also had similar side effects, I see no major harm in this change.

### Checklist
- [ ] I have commented my code
- [ ] I have written javadoc (required for public methods)


oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -485,61 +487,57 @@ private static Polygon constructMultipolygonPart(
...
-    // make a (mutable) copy of the polygons array
+    // make a (mutable) copy of the lines list
     List<LinkedList<OSMNode>> ways = lines
         .map(line -> line.collect(Collectors.toCollection(LinkedList::new)))
         .filter(nodesList -> !nodesList.isEmpty())
         .collect(Collectors.toCollection(LinkedList::new));
 
     List<LinkedList<OSMNode>> joined = new LinkedList<>();
+    // iterate until there are no more ways left to process
     while (!ways.isEmpty()) {
```
joker234, line 498: Is it theoretically possible, that `ways` never gets empty and stays in an infinite loop? I couldn't find any `break` or `return` in this loop. Normally, this should never be an issue, because of the next line.

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/OSHDBGeometryBuilder.java
```diff
@@ -420,6 +420,8 @@ private static void cutAtSegment(LinkedList<OSMNode> ring, Segment cutSegment) {
         ring.add(ring.removeFirst());
       }
     }
+    assert false : "cut segment not found in ring";
+    throw new RuntimeException("cut segment not found in ring");
```
joker234, line 424: Is this something you want to address? _Define and throw a dedicated exception instead of using a generic one._

oshdb-util/src/test/java/org/heigit/ohsome/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java
```diff
@@ -579,12 +579,11 @@ public void test745() {
 
   @Test
   public void test746() {
-    // IImpossible multipolygon out of two way.
+    // Impossible multipolygon out of two ways.
     OSMEntity entity1 = testData.relations().get(746900L).get(0);
     try {
       Geometry result = OSHDBGeometryBuilder.getGeometry(entity1, timestamp, tagInterpreter);
       assertTrue(result instanceof GeometryCollection || result instanceof Polygonal);
-      assertEquals(2, result.getNumGeometries());
```
joker234, line 587: What is the expected outcome of this test? Shouldn't we test for a proper result? Or is this already covered by another test?

oshdb-util/src/test/java/org/heigit/ohsome/oshdb/util/geometry/relations/OSHDBGeometryBuilderMultipolygonInvalidInnersTest.java
```diff
@@ -25,10 +25,18 @@ public OSHDBGeometryBuilderMultipolygonInvalidInnersTest() {
   }
 
   @Test
-  public void test() {
+  public void testDuplicateInnerRings() {
     // data has invalid (duplicate) inner rings
     OSMEntity entity = testData.relations().get(1L).get(0);
     Geometry result = OSHDBGeometryBuilder.getGeometry(entity, timestamp, tagInterpreter);
     assertTrue(result instanceof Polygon);
```
joker234, line 32: Is it sufficient to test for some polygon? Don't we have to test if it is the polygon we would expect as result? e.g. the polygon in `invalid-inner-rings.osm` with or without the inner hole. Or is this already covered by another test?

oshdb-util/src/test/java/org/heigit/ohsome/oshdb/util/geometry/relations/OSHDBGeometryBuilderMultipolygonInvalidInnersTest.java
```diff
@@ -25,10 +25,18 @@ public OSHDBGeometryBuilderMultipolygonInvalidInnersTest() {
...
   }
+
+  @Test
+  public void testTouchingIncompleteInnerRings() {
+    // data has invalid (duplicate) inner rings
+    OSMEntity entity = testData.relations().get(2L).get(0);
+    Geometry result = OSHDBGeometryBuilder.getGeometry(entity, timestamp, tagInterpreter);
+    assertTrue(result instanceof Polygon);
```
joker234, line 40: Same question here. What would we expect with an invalid second inner “ring” and a valid ring? Or is this already covered by another test?

joker234:
Is the Milestone 0.7 fitting to this issue even tough we want to fix that directly for the 0.6 version?

tyrasd:
> Is the Milestone 0.7 fitting to this issue even tough we want to fix that directly for the 0.6 version?

I wasn't super sure about that either, but adding it to the 0.6 milestone seemed "wrong" because it's already closed; maybe not adding it to any milestone would be better? My thinking was: ok, we fix this for the current next version (0.7, indicated by the milestone) and additionally backport it as well (to 0.6, indicated by the `priority:high` label). :thinking:

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/OSHDBGeometryBuilder.java
tyrasd: yeah, on each iteration, at least one element gets removed from the list (line 499). so the loop will definitely eventually be done. I tried to write this a bit differently to make it more obvious (e.g. by use of a `for (…; …; …)` loop, but in this particular situation it wasn't very practical/helpful.

oshdb-util/src/test/java/org/heigit/ohsome/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java
tyrasd: The point is that in all tests with invalid input data, we basically don't care what _exactly_ comes out as the result, as long as it is anything _reasonable_ (here: not `null`, either polyon or geometry collection) and no exceptions are thrown.

Not sure why the `numGeometries` was tested here before – I mean, it makes some sense if the result is a GeometryCollection (as was returned before this PR), but if one decides to build a Polygon out of the invalid input data, it's kinda hard to construct something with two parts. :innocent:

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/OSHDBGeometryBuilder.java
tyrasd: Right. I replace it with an `IllegalStateException`.

oshdb-util/src/test/java/org/heigit/ohsome/oshdb/util/geometry/relations/OSHDBGeometryBuilderMultipolygonInvalidInnersTest.java
tyrasd: There should already be "enough" tests covering the basic building of multipolygons with various configurations of inner rings (e.g. see `test750` ff.). So here we only test for the algorithm not choking up on this particular invalid input data.

oshdb-util/src/test/java/org/heigit/ohsome/oshdb/util/geometry/relations/OSHDBGeometryBuilderMultipolygonInvalidInnersTest.java
tyrasd: see https://github.com/GIScience/oshdb/pull/343#discussion_r565943183

IMHO, we shouldn't care. There would be so many possible ways to handle invalid data, in principle. Sure, some perhaps "better" ways and some worse ways, but if we restrict us too much here we loose quite some flexibility in the implementation. And I can live with some ambiguity in the output if the input is not really valid in the first place.

joker234 (approved):

joker234:
> > Is the Milestone 0.7 fitting to this issue even tough we want to fix that directly for the 0.6 version?
> 
> I wasn't super sure about that either, but adding it to the 0.6 milestone seemed "wrong" because it's already closed; maybe not adding it to any milestone would be better? My thinking was: ok, we fix this for the current next version (0.7, indicated by the milestone) and additionally backport it as well (to 0.6, indicated by the `priority:high` label). :thinking:

I was thinking about not adding it to any milestone would be okay. But keeping it in 0.7 is okay as well.

PR #346, opened 2021-01-28
update main documentation to latest release

tyrasd:
addendum to #344 (and perhaps also #339)


joker234 (approved):
Shouldn't you update this number as well? https://github.com/GIScience/oshdb/blame/0c95271/CHANGELOG.md#L22 Maybe you can write `0.6` instead and drop the patch version number.

tyrasd:
> Shouldn't you update this number as well? https://github.com/GIScience/oshdb/blame/0c95271/CHANGELOG.md#L22 Maybe you can write `0.6` instead and drop the patch version number.

makes sense :+1:

joker234 (approved):

Issue #214, opened 2020-02-03
switch to Martinez-Rueda polygon clipping algorithm


tyrasd:
Paper: https://www.sciencedirect.com/science/article/pii/S0965997813000379

apparently, this one is significantly faster than other algorithms (maybe incl. JTS's one?) and can better handle special cases (like self-intersecting polygons).

Hagellach37:
Out of scope for OSHDB.

PR #362, opened 2021-03-22
catch and handle topology exceptions when splitting geometries

tyrasd:
### Description

After #272 (version `0.6.0`), topology exceptions thrown during the `intersection` calculations in the `GeometrySplitter` class are not (always) catched anymore. This solves the problem by adding an additional try-catch block and returning an empty geometry if a topology exception is occurring.

### Checklist

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/GeometrySplitter.java
+++ b/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/GeometrySplitter.java
@@ -22,6 +22,7 @@ import org.heigit.ohsome.oshdb.util.geometry.fip.FastBboxInPolygon;
 import org.heigit.ohsome.oshdb.util.geometry.fip.FastBboxOutsidePolygon;
 import org.heigit.ohsome.oshdb.util.geometry.fip.FastPolygonOperations;
 import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.Polygonal;
 import org.locationtech.jts.geom.TopologyException;
 import org.locationtech.jts.geom.prep.PreparedGeometry;
@@ -118,14 +119,17 @@ class GeometrySplitter<U extends Comparable<U>> implements Serializable {
           try {
             boolean intersects = pg.intersects(snapshotGeometry);
             if (!intersects) {
-              return Stream.empty(); // not actually intersecting -> skip
+              // not actually intersecting -> skip
+              return Stream.empty();
             } else {
               return Stream.of(new IndexData<>(index, new OSMEntitySnapshot(data,
-                  new LazyEvaluatedObject<>(() -> poop.intersection(snapshotGeometry))
+                  new LazyEvaluatedObject<>(() ->
+                      faultTolerantIntersection(snapshotGeometry, poop))
               )));
             }
           } catch (TopologyException ignored) {
-            return Stream.empty(); // JTS cannot handle broken osm geometry -> skip
+            // JTS cannot handle broken osm geometry -> skip
+            return Stream.empty();
           }
         }).collect(Collectors.toMap(IndexData::getIndex, IndexData::getData));
   }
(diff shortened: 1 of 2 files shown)


FabiKo117:
We used to say that with the OSHDB the user has access to erroneous OSM data as well, including features that were falsely tagged, or also had a broken geometry. How would this be affected with this fix here? Could someone still perform an analysis looking at such data that has/had some issues regarding its topology through the OSHDB?

tyrasd:
> We used to say that with the OSHDB the user has access to erroneous OSM data

I'm not sure where exactly and in which context we said that, but certain data problems are just too grave to be able to access via the `oshdb-api` – entities with no or invalid coordinates, for example. The issue with topology exceptions is that these come from an external library (JTS), which the `oshdb-api` needs for performing geometry operations like clipping. When JTS refuses to do that (because the OSM data is too broken), there's little alternatives at the moment for us. I once proposed to evaluate a different algorithm for clipping in #214, which might be more robust, but that's definitely out of scope of this PR.

What I think we always promised is that the `oshdb-api` would not crash, even when confronted with broken/invalid OSM data, see e.g. [our unit test](https://github.com/GIScience/oshdb/blob/5e444f11da09b504e3596427d374243b91f12f0c/oshdb-util/src/test/java/org/heigit/ohsome/oshdb/util/geometry/osmtestdata/OSHDBGeometryBuilderTestOsmTestData7xxTest.java#L248). This PR simply restores the behavior from version 0.5 again.

> Could someone still perform an analysis looking at such data that has/had some issues regarding its topology through the OSHDB?

Yes, for example by using a big enough bounding box, so that clipping doesn't affect the elements one is interested in. Or by using a custom `TagInterpreter`, or by using the low level data access APIs of the OSHDB instead of the `oshdb-api`.

FabiKo117:
I see.. and yes of course we can only deal with what JTS can deal with in that matter. The workaround with just using a bigger bbox would already be sufficient I guess.

Do we potentially have to add another test or extend one of the existing ones to also cover this specific issue? Because you say that it was something that was working as intended in v0.5.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/GeometrySplitter.java
```diff
@@ -197,19 +201,32 @@ D getData() {
...
           }
         }).collect(Collectors.toMap(IndexData::getIndex, IndexData::getData));
   }
 
+  private static Geometry faultTolerantIntersection(Geometry subject, FastPolygonOperations poop) {
+    try {
+      return poop.intersection(subject);
+    } catch (TopologyException ignored) {
+      // JTS cannot handle broken osm geometry -> return empty geometry
+      return (new GeometryFactory()).createGeometryCollection();
```
joker234, line 226: I assume it's irrelevant which `Geometry` you use here? Or is there a specific reason you use `GeometryCollection`?

joker234 (approved):
looks good to me. Only the question by @FabiKo117 is still an open question I would be interested in:
> Do we potentially have to add another test or extend one of the existing ones to also cover this specific issue? Because you say that it was something that was working as intended in v0.5.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/GeometrySplitter.java
tyrasd: I also thought about using the geometry type of `subject` to return a matching empty geometry: mostly/only(?) empty polygon or empty geometry collection, but am not sure if it's worth the additional complexity. My reasoning for choosing the most GeometryCollection type is that broken polygons could also end up as geometry collections if our geometry building algorithm was a bit smarter to detect invalid geometries and used the same [fall back](https://github.com/GIScience/oshdb/blob/master/oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/OSHDBGeometryBuilder.java#L136) as it does for other error cases.

tyrasd:
> Do we potentially have to add another test or extend one of the existing ones to also cover this specific issue?

Unfortunately, it's hard to find data which triggers this bug in our limited test data set. :grimacing:

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/GeometrySplitter.java
joker234: thanks for the explanation :+1:

PR #363, opened 2021-03-23
avoid tag-translating in filter optimizer, fixes crash when filters contains a non-existent tag 

tyrasd:
### Description
A oshdb filter goes through some optimization steps for increased performance, e.g. for simple `key=value` queries, it will use a more efficient code path than for example a filter like `key1=value1 or key2=value2`. The "fast" path crashed, when a non-existent tag is contained in the filter: The code tries to convert the tag from the OSHDB-internal representation back to strings. This normally works fine, but not for tags which don't exist.

This fixes the bug by replacing the conversion from tag ids to strings by a newly added private method for using tag ids directly for filtering.

This PR also fixes a few minor typos (`entitiy` -> `entity`) and refactors some of the `osmTag` filters slightly (to also use the new helper method and be more similar to each other).

### Corresponding issue
Closes https://github.com/GIScience/ohsome-api/issues/154

### Checklist

### DIFF
--- a/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
+++ b/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
@@ -414,6 +414,10 @@ public abstract class MapReducer<X> implements
   @Deprecated
   @Contract(pure = true)
   public MapReducer<X> osmType(Set<OSMType> typeFilter) {
+    return osmTypeInternal(typeFilter);
+  }
+
+  private MapReducer<X> osmTypeInternal(Set<OSMType> typeFilter) {
     MapReducer<X> ret = this.copy();
     typeFilter = Sets.intersection(ret.typeFilter, typeFilter);
     if (typeFilter.isEmpty()) {
@@ -2155,7 +2159,7 @@ public abstract class MapReducer<X> implements
       OSHDBTagKey key = ((TagFilterEqualsAny) filter).getTag();
       return mapRed.osmTag(key);
     } else if (filter instanceof TypeFilter) {
-      return mapRed.osmType(((TypeFilter) filter).getType());
+      return mapRed.osmTypeInternal(EnumSet.of(((TypeFilter) filter).getType()));
     } else if (filter instanceof AndOperator) {
       return optimizeFilters0(optimizeFilters0(mapRed,
           ((AndOperator) filter).getLeftOperand()),
--- a/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
+++ b/oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
@@ -22,7 +22,6 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.function.DoubleUnaryOperator;
-import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
```diff
@@ -2143,18 +2150,13 @@ private void checkTimeout() {
...
-      OSMTagKey key = this.getTagTranslator().getOSMTagKeyOf(
-          ((TagFilterEqualsAny) filter).getTag());
+      return mapRed.osmTag(((TagFilterEquals) filter).getTag());
+    } else if (filter instanceof TagFilterEqualsAny) {
+      OSHDBTagKey key = ((TagFilterEqualsAny) filter).getTag();
       return mapRed.osmTag(key);
-    }
-    if (filter instanceof TypeFilter) {
+    } else if (filter instanceof TypeFilter) {
       return mapRed.osmType(((TypeFilter) filter).getType());
```
joker234, line 2158: The usage of `osmType` seems to be deprecated here.

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOSHDBFilter.java
```diff
@@ -115,4 +116,20 @@ public void testFilterGroupByEntity() throws Exception {
...
     assertEquals(osmTypeFilterResult, stringFilterResult);
   }
+
+  @Test
+  @SuppressWarnings("ResultOfMethodCallIgnored")
+  public void testFilterNonExistentTag() throws Exception {
+    FilterParser parser = new FilterParser(new TagTranslator(oshdb.getConnection()));
+    try {
+      createMapReducerOSMEntitySnapshot()
+          .filter(parser.parse("type:way and nonexistentkey=*"))
```
joker234, line 126: Is it good to use something which people could actually use, like `nonexistentkey`? Or would something “randomly” generated be better? I assume you checked it with the test database, though. At least `nonexistent` exists in the current OSM dataset ;)

joker234 (commented):
So basically, apart from some reformatting and prettyfying the code, the change is to use `OSHDBTag` and `OSHDBTagKey` directly instead of translating it to `OSMTag` or `OSMTagKey` objects first? This prevents the translator to fail and we can return an empty response?

tyrasd:
> So basically […] the change is to use `OSHDBTag` and `OSHDBTagKey` directly instead of translating it to `OSMTag` or `OSMTagKey` objects first? This prevents the translator to fail and we can return an empty response?

Exactly.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapReducer.java
tyrasd: right. But the deprecation is very likely not going to remove the `osmType` method, but only make it `private`. I didn't find a good way to represent this kind of deprecation with the `@Deprecated` annotation… :thinking:  Maybe we should add a private, non-deprecated `osmTypeInternal` and use that one both here and from the to-be-removed `osmType` method?

joker234: We don't have to do that now in this PR.

oshdb-api/src/test/java/org/heigit/ohsome/oshdb/api/tests/TestOSHDBFilter.java
tyrasd: > At least nonexistent exists in the current OSM dataset

:rofl: 

Actually, I didn't check it, but made sure that the test fails before applying the fix. I guess it's ok to leave it as is, since we can control the test data used for the unit tests: The keytables of the test data set only contain a handful of tags at the moment.

joker234 (dismissed):

joker234 (approved):

PR #369, opened 2021-04-12
refactor oshdb core (and more)

rtroilo:
Refactoring oshdb core. This PR touches a lot of files because of moving of classes to different packages.

## renaming methods
- OSMWay.getRef() to OSMWay.getMembers()
this makes is more unified with the OSMRelation interface
- OSHDBTimestamp.getRawUnixTimestamp() to OSHDBTimestamp.getEpochSeconds()
 
## deprecations of methods
- OSMEntity.getRawTag
- OSHEntity.getRawTagKeys

## removed methods
- OSHDBBoundingBox.add/set
 this makes OSHDBBoundingBox immutable!

## reorganisation
 - move oshdb.util.OSHDBTag to oshdb.OSHDBTag
 - move oshdb.util.OSHDBTimestamp to oshdb.OSHDBTimestamp
 - move oshdb.util.OSHDBBoundingBox to oshdb.OSHDBBoundingBox
 - move oshdb.util.CellIterator.OSHEntityFilter to oshdb.osh.OSHEntityFilter
 - move oshdb.util.CellIterator.OSMEntityFilter to oshdb.osm.OSMEntityFilter
 - move oshdb-api.generic.function to oshdb.util.function
 makes a common/central place for our extended java function api.

## new interfaces
- OSHDBTimeable
 main interface for OSHDBTimestamp, allows to "flatten" OSMEntity and reduce object creation.
 add methods isAfter/isBefore to OSMEntity so that we directly can check like `osmEntity.isAfter(timestamp)`
- OSHDBBoundable
 main interface for OSHDBBoundingBox allows to "flatten" OSHEntity and reduce object creation.
 add methods intersects/isInside to OSHEntity so that we direcly can check like `oshEntity.intersects(bbox)`

### notes
- code-style fixes will be in a separate PR to not clutter this PR with "unrelated" changes 

### Checklist
- [x] I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/issues/new) in the corresponding repository -> https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/merge_requests/15
- [x] I have adjusted the [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-benchmarks/-/issues/new) in the corresponding repository -> https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-benchmarks/-/merge_requests/14


oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/MapAggregator.java
```diff
@@ -748,22 +747,8 @@ public void forEach(SerializableBiConsumer<U, List<X>> action) throws Exception
...
-      @Override
-      public X getValue() {
-        return d.getValue();
-      }
-
-      @Override
-      public X setValue(X value) {
-        throw new RuntimeException("cannot modify the value of this entry");
-      }
-    });
```
rtroilo, line 766: I replaced the anonym class with AbstractMap.SimpleImmutableEntry, I thought this makes sense in this context.

tyrasd:
> * move oshdb-api.generic.function to oshdb.util.function
> makes a common/central place for our extended java function api.

Not sure I can follow here. AFAICS, the "extended" java functions are only ever used in the oshdb-api module, and are only necessary elsewhere (we only need it when running queries on a remote ignite cluster with peer class loading enabled). IMHO we should put as little "stuff" into the "core" module as possible. Keeping it slim allows that some fixes don't need to be deployed on the ignite cluster for example.

rtroilo:
I saw a commit https://github.com/GIScience/oshdb/commit/46ec2eb309c964c5b3c6b8e0ac810e548b3f94b8 which introduces an other set of PredicateSerializables again, because of circular dependencies.
So to solve this, and to avoid to have incompatible PredicateSerializables I moved them to a more central place. Also the OSH/OSMEntityFilter fits nicely into the osh/osm packages.

rtroilo:
> IMHO we should put as little "stuff" into the "core" module as possible.

Yes sounds good, so what do you think about to move the SerializedFunctions into the oshdb-util module oshdb.util.function instead of the oshdb-api module. Because there are also used by the oshdb.util.celliterator.CellIterator and oshdb.filter.Filter

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/celliterator/CellIterator.java
```diff
@@ -49,6 +48,8 @@
 import org.locationtech.jts.geom.TopologyException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Streams;
```
tyrasd, line 52: odd… those imports should have remained at the top (before `import java.io.Serializable;`).

tyrasd: (similar issues exist also in other classes, .g. `Transform*.java`)

rtroilo: so I reverted the reordering of the import statement. Is there still something missing before we can merge this PR?

tyrasd: at first glance no, but a code review is definitely still missing :stuck_out_tongue_closed_eyes: can you remove the "draft" status, then I'll take a look.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundingBox.java
```diff
@@ -75,52 +65,35 @@ public OSHDBBoundingBox(double minLon, double minLat, double maxLon, double maxL
...
-  }
-
-  public double getMinLon() {
-    return minLon * OSHDB.GEOM_PRECISION;
+  public OSHDBBoundingBox add(OSHDBBoundingBox bbox){
+    return new OSHDBBoundingBox(
+        Math.min(minLon,bbox.minLon),
+        Math.min(minLat,bbox.minLat),
+        Math.max(maxLon,bbox.maxLon),
+        Math.max(maxLat,bbox.maxLat));
```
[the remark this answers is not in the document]
tyrasd: PS: similarly, `OSHDBBoundingBox::intersect` could also be renamed to `intersection` made non-static, have only one parameter `OSHDBBoundable other` and live in the interface.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
```diff
@@ -0,0 +1,52 @@
+package org.heigit.ohsome.oshdb;
+
+public interface OSHDBBoundable {
```
tyrasd, line 3: if we're touching this topic:

Why don't we go ahead and change the internal coordinate representation from the 64 bit longs to 32 bit integers: As long as `GEOM_PRECISION` is not bigger than 1E7, 32 bits are sufficient to store coordinates.

This would leave the question how we want to name the corresponding getters for the internal representation (`int getMinLonLong()` would not make much sense IMO). Maybe `int getMinLonInternal` could be an option?

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
```diff
@@ -0,0 +1,52 @@
...
+    return (otherBbox != null) 
+        && (getMinLatLong() >= otherBbox.getMinLatLong()) 
+        && (getMaxLatLong() <= otherBbox.getMaxLatLong())
+        && (getMinLonLong() >= otherBbox.getMinLonLong()) 
+        && (getMaxLonLong() <= otherBbox.getMaxLonLong());
+  }
+  
+  default boolean isPoint() {
+    return getMinLonLong() == getMaxLonLong() && getMinLatLong() == getMaxLatLong();
+  }
```
tyrasd, line 51: missing javadoc for these methods

```suggestion
  /** Returns true if the {@code other} boundable object intersects this. */
  default boolean intersects(OSHDBBoundable other) {
    return (other != null)
        && (getMaxLatLong() >= other.getMinLatLong())
        && (getMinLatLong() <= other.getMaxLatLong())
        && (getMaxLonLong() >= other.getMinLonLong())
        && (getMinLonLong() <= other.getMaxLonLong());
  }

  /** Returns true this boundable is inside the {@code other} object. */
  default boolean isInside(OSHDBBoundingBox other) {
    return (other != null) 
        && (getMinLatLong() >= other.getMinLatLong())
        && (getMaxLatLong() <= other.getMaxLatLong())
        && (getMinLonLong() >= other.getMinLonLong())
        && (getMaxLonLong() <= other.getMaxLonLong());
  }

  /** Returns true this boundable is degenerated to a point. */
  default boolean isPoint() {
    return getMinLonLong() == getMaxLonLong() && getMinLatLong() == getMaxLatLong();
  }
```

PS: the logic of `isInside` could be reversed and be called `contains`, to be similar to the [JTS](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/Geometry.html#contains-org.locationtech.jts.geom.Geometry-) convention.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
```diff
@@ -0,0 +1,52 @@
+package org.heigit.ohsome.oshdb;
+
+public interface OSHDBBoundable {
+
```
tyrasd, line 4: AFAICS, this empty line (and quite a few more in this file and perhaps the whole PR) contains two spaces. Empty lines should be fully empty. In general, there should not be any trailing whitespace at the end of any line.

image

oshdb/CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+ - CellIterator.OSHEntityFilter/OSMEntityFilter - oshdb.osh.OSHEntityFilter, oshdb.osm.OSMEntityFilter
+ - oshdb-api.generic.function - oshdb.util.function
+ 
+
+### breaking changes
+
+### new features
+ - new interface OSHDBTimeable
+ - new interface OSHDBBoundable
+
```
tyrasd, line 22: ehrm. not sure if I like the separate per-module changelog. Why not just include this in the general CHANGELOG (perhaps slightly condensed)? Especially since this includes breaking changes (btw, aren't moved and renamed classes/methods are **breaking changes** since people do need to adjust their code accordingly).

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTag.java
```diff
@@ -27,15 +28,15 @@ public boolean isPresentInKeytables() {
 
   @Override
   public boolean equals(Object o) {
-    return o instanceof OSHDBTag &&
-        ((OSHDBTag)o).key == this.key && ((OSHDBTag)o).value == this.value;
+    return o instanceof OSHDBTag
```
tyrasd, line 31: superfluous trailing whitespace

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTag.java
```diff
@@ -1,12 +1,13 @@
-package org.heigit.ohsome.oshdb.util;
+package org.heigit.ohsome.oshdb;
 
 import java.io.Serializable;
 import java.util.Objects;
 
 public class OSHDBTag implements Serializable {
+
```
tyrasd, line 7: empty line should be empty

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTag.java
```diff
@@ -27,15 +28,15 @@ public boolean isPresentInKeytables() {
...
+    return o instanceof OSHDBTag 
+        && ((OSHDBTag) o).key == this.key && ((OSHDBTag) o).value == this.value;
   }
 
   @Override
   public int hashCode() {
     return Objects.hash(this.key, this.value);
   }
-
+
```
tyrasd, line 39: empty line should be empty

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTimeable.java
```diff
@@ -0,0 +1,40 @@
...
+  
+  default OSHDBTimestamp getTimestamp() {
+    return new OSHDBTimestamp(getEpochSecond());
+  }
+  
+  default boolean isBefore(OSHDBTimeable other) {
+    return getEpochSecond() < other.getEpochSecond();
+  }
+  
+  static boolean isBefore(OSHDBTimeable a, OSHDBTimeable b) {
```
tyrasd, line 20: are this static method and the corresponding `static boolean isAfter` really needed?

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTimeable.java
```diff
@@ -0,0 +1,40 @@
...
+  
+  default boolean isAfter(OSHDBTimeable other) {
+    return getEpochSecond() > other.getEpochSecond();
+  }
+  
+  static boolean isAfter(OSHDBTimeable a, OSHDBTimeable b) {
+    return a.isAfter(b);
+  }
+  
+  static int compare(OSHDBTimeable a, OSHDBTimeable b) {
```
tyrasd, line 32: why not make the class implement `Comparable<OSHDBTimeable>` instead?

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTimeable.java
```diff
@@ -0,0 +1,40 @@
...
+    return Long.compare(a.getEpochSecond(), b.getEpochSecond());
+  }
+  
+  static String toString(OSHDBTimeable o) {
+    ZonedDateTime zdt = ZonedDateTime.ofInstant(Instant.ofEpochSecond(o.getEpochSecond()), ZoneOffset.UTC);
```
tyrasd, line 37: this line is longer than 100 characters

tyrasd (changes requested):
_For this review I only looked at some of the code (i.e. where there are comments below)._

In general I like the idea behind the introduction of the new `OSHDBTimeable` and `OSHDBBoundable` interfaces, but the naming could be "improved" maybe? IMHO, it could be made clearer the the time aspect is only for points in time, while the spatial aspect is (in general) for bounded regions.

Maybe the point-in-time interface could be just `OSHDBEvent` (or `OSHDBTemporal`) which would leave us the door open to later introduce a class for time-intervals similar to `OSHDBBoundable`) without confusion?!

PS: I found that there were a lot of lines introduced in this PR which contain trailing spaces or empty lines consisting of only spaces. IMHO this should be avoided. Is it not possible to configure your code editor to automatically remove trailing whitespace on save?

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
rtroilo: I like to name the methods closer to JTS. So I would rename 
- isInside to  coveredBy (even this method actually not exist in JTS Envelope but to not change the logic on the call sites)
and
- add a new method covers which is then the reverse of (isInside/coveredBy)

Thank you for your comment suggestion, I will take them and extend them to be full javadoc comments.
Like
```
  /**
   * Returns true if the {@code other} {@code OSHDBBoundable} intersects this.
   * 
   * @param other  the <code>OSHDBBoundable</code> which this <code>OSHDBBoundable</code> is
   *          being checked for intersecting
   * @return <code>true</code> if the <code>OSHDBBoundable</code>s intersect
   */
  default boolean intersects(OSHDBBoundable other) {
```

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
rtroilo: I just check, within the oshdb there is really only one place where we actually access the double return value variants.
`OSHDBGeometryBuilder::getGeometry`

So my suggestion here is, to remove the double return value methods completely and rename/change the long return value variants to 
```
int minLon();
int minLat();
int maxLon();
int maxLat();
 ```
 This would make it clear, that the OSHDBBoundable/OSHDBoundingBox use OSM-Integer coordinates. If the user needs the double coordinates back, she has to use converter methods like `OSHDBGeometryBuilder::getGeometry/getEnvelope`

File: oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundingBox.java [hunk 14]
```diff
@@ -75,52 +65,35 @@ public OSHDBBoundingBox(double minLon, double minLat, double maxLon, double maxL
...
-  public double getMaxLat() {
-    return maxLat * OSHDB.GEOM_PRECISION;
-  }
-  
   public long getMaxLatLong(){
     return maxLat;
   }
+  
+  @Override
+  public OSHDBBoundingBox getBoundingBox() {
```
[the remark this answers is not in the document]
rtroilo: Actually it might be usefull/needed. 
The `OSHDBBoundable::getBoundingBox()` returns an instance of `OSHDBBoundingBox` and not `OSHDBBoundable`. 
So if you want to store the bboxes for example into a map it makes a different if you use `OSHDBBoundingBox` or `OSHDBBoundable`
e.g.
```
 entity a -> bbox(0,0,1,1)
 entity b -> bbox(0,0,1,1)
 boundable ba = (OSHDBBoundable) a
 boundable bb = (OSHDBBoundable) b
 
 ba.equals(bb) -> false
 a.getBoundingBox().equals(b.getBoundingBox()) -> true
```

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
tyrasd: sounds good. :+1:

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundingBox.java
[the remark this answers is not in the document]
tyrasd: Ok, I see. But as we don't need it at the moment, it should be omitted for now, shouldn't it?

Also, a slightly less confusing alternative solution could be to add a copy-constructor to the `OSHDBBoundingBox` class:

```java
  public OSHDBBoundingBox(OSHDBBoundable boundable) {
    this.minLon = boundable.getMinLonLong();
    …
```

then you could do something like this:

```
 entity a -> bbox(0,0,1,1)
 OSHDBBoundingBox ba = new OSHDBoundingBox(a)
 OSHDBBoundingBox bb = …
 
 ba.equals(bb) -> true
```

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTimeable.java
tyrasd: :arrow_up: this still waits for an answer. also, didn't we say we want to rename this interface (see https://github.com/GIScience/oshdb/pull/369#pullrequestreview-641250682)?

tyrasd:
PS: I've crated branches to update the examples and benchmarks repositories:

* https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/merge_requests/15
* https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-benchmarks/-/merge_requests/14

tyrasd:
> * [ ] I have written javadoc (required for public methods)
> * [ ] I have added sufficient unit tests

It would be nice to check these off, at least for the newly created interfaces. :pray:

oshdb/src/main/java/org/heigit/ohsome/oshdb/impl/osh/OSHNodeImpl.java
```diff
@@ -73,16 +78,45 @@ public static OSHNodeImpl instance(final byte[] data, final int offset, final in
...
+          maxLon = Math.max(maxLon, osm.getLon());
+
+          minLat = Math.min(minLat, osm.getLat());
+          maxLat = Math.max(maxLat, osm.getLat());
+        }
+      }
+
+      if (minLon == Long.MAX_VALUE || minLat == Long.MAX_VALUE) {
+        minLon = minLat = -1;
+        maxLon = maxLat = 1;
```
tyrasd, line 112: what does this do?

1. the variables `minLon` etc. are set here, but never used. see [spotbugs warning](https://jenkins.ohsome.org/job/oshdb/view/change-requests/job/PR-369/27/spotbugs/moduleName.736606877/packageName.-1331010043/fileName.1096615460/category.79242641/source.35911c19-2c3c-4bc7-b62c-1240086c5e55/#112)
2. did you mean to set `this.minLon` (etc.) to these values? if yes, I don't necessarily see how setting it to a `(-1,-1,1,1)` bbox around null island makes much sense… can you elaborate please?

oshdb/src/main/java/org/heigit/ohsome/oshdb/impl/osh/OSHNodeImpl.java
```diff
@@ -73,16 +78,45 @@ public static OSHNodeImpl instance(final byte[] data, final int offset, final in
...
       final long baseTimestamp, final long baseLongitude, final long baseLatitude,
-      final byte header, final long id, final OSHDBBoundingBox bbox, final int[] keys,
+      final byte header, final long id, long minLon, long minLat, long maxLon, long maxLat, 
+      final int[] keys,
       final int dataOffset, final int dataLength) {
     super(data, offset, length, baseNodeId, baseTimestamp, baseLongitude, baseLatitude, header, id,
-        bbox, keys, dataOffset, dataLength);
+        minLon, minLat, maxLon, maxLat, keys, dataOffset, dataLength);
+    
+    // correct bbox!
```
tyrasd, line 94: could you please state why this correcting is necessary? shouldn't this be even a case for throwing an exception (i.e. when the method is called with wrong parameters)?

rtroilo: this is a remain of the current logic, as nodes don't necessary have a bbox(but a point in case of a single version) the bbox information is retrieved, by checking the versions it self.
The idea behind this is/was, to save space by not duplicating information.
But this code will be reworked soon.

oshdb/src/main/java/org/heigit/ohsome/oshdb/impl/osh/OSHNodeImpl.java
rtroilo: good catch!

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTimeable.java
rtroilo: OSHDBTimeable now called OSHDBTemporal

>    why not make the class implement Comparable instead?

 you mean the interface OSHDTimeable/OSHDBTemporal should extend Comparable right? This is actually quite problematic, because all implementations of this interface would then inherit the Comparable which is in most case not the desired behavior. The reason for the static compare is to make it more convenient to use it as a Comparator function in e.g. TreeMap/Sets. We can leave it in the interface or remove it.

oshdb/CHANGELOG.md
rtroilo: this is done, isn't it? 
removed the changelog from oshdb module and added it to oshdb-parent changelog.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
rtroilo: apparently our checkstyle ruleset does not complain about not empty blank lines. Maybe you should add this to our rules, so that it becomes more visible.
This should also be fixed.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTag.java
rtroilo: should be fixed, but we should also add this as a rule to our checkstyle. So that it becomes more visible.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
rtroilo: I will address this issue, changing long coordinates to integer, in a following up PR which include also the OSMCoordinates and conversions from (int/long to double) from PR #320

rtroilo:
I removed the getTimestamp for the OSHDBTemporal interface as well.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBTimeable.java
tyrasd: > […] because all implementations of this interface would then inherit the Comparable which is in most case not the desired behavior. The reason for the static compare is to make it more convenient to use it as a Comparator function in e.g. TreeMap/Sets. We can leave it in the interface or remove it.

OK, that makes sense. Let's leave it as is then.

oshdb/CHANGELOG.md
tyrasd: yes!

rtroilo:
> > * [ ]  I have written javadoc (required for public methods)
> > * [ ]  I have added sufficient unit tests
> 
> It would be nice to check these off, at least for the newly created interfaces. pray

I added the javadoc but I'm not sure about the uni-tests for the interfaces. I think the should be covered by the implementing classes. Do you think that is enough, or should I write some test again.

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
tyrasd: yes, because strictly speaking the java rules don't specify this as a rule (while the same style guides recommend to trim trailing whitespace for other languages such as C#, see https://github.com/google/styleguide/issues/262). but yes, we should think about adding it to our own rules. I will open a ticket for that right now.

rtroilo: Sounds good, thank you. As I said this makes non empty lines or trailing whitespaces more prominent to remove :+1:

oshdb/src/main/java/org/heigit/ohsome/oshdb/OSHDBBoundable.java
tyrasd: OK

tyrasd (approved):

rtroilo:
Thank you and good review!

PR #372, opened 2021-04-20
configure sonarcloud

tyrasd:
### Description
sets up sonarcloud static code analysis: https://sonarcloud.io/dashboard?id=org.heigit.ohsome%3Aoshdb

### New or changed dependencies
- update parent to version 2.9

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,6 +61,11 @@ pipeline {
     stage ('Reports and Statistics') {
       steps {
         script {
+          // START CUSTOM oshdb
+          withSonarQubeEnv('sonarcloud GIScience/ohsome') {
+            sh "mvn sonar:sonar -Dsonar.branch.name=${env.BRANCH_NAME} -Dsonar.projectName=OSHDB"
+          }
+          // END CUSTOM oshdb
           report_basedir = "/srv/reports/${REPO_NAME}/${VERSION}_${env.BRANCH_NAME}/${env.BUILD_NUMBER}_${LATEST_COMMIT_ID}"
 
           // jacoco
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.heigit.ohsome</groupId>
     <artifactId>ohsome-parent</artifactId>
-    <version>2.8</version>
+    <version>2.9</version>
   </parent>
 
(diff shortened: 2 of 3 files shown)


Jenkinsfile
```diff
@@ -61,6 +61,11 @@ pipeline {
     stage ('Reports and Statistics') {
       steps {
         script {
+          withSonarQubeEnv('sonarcloud GIScience/ohsome') {
+            // START CUSTOM oshdb
+            sh "mvn sonar:sonar -Dsonar.branch.name=${env.BRANCH_NAME} -Dsonar.projectName=OSHDB"
+            // END CUSTOM oshdb
+          }
```
joker234, line 68:
```suggestion
          // START CUSTOM oshdb
          withSonarQubeEnv('sonarcloud GIScience/ohsome') {
            sh "mvn sonar:sonar -Dsonar.branch.name=${env.BRANCH_NAME} -Dsonar.projectName=OSHDB"
          }
          // END CUSTOM oshdb
```

pom.xml
```diff
@@ -46,6 +52,8 @@
     <roaringbitmap.version>0.6.59</roaringbitmap.version>
     <rxjava2.version>2.1.9</rxjava2.version>
     <slf4j.version>1.7.25</slf4j.version>
+    <sonar.moduleKey>${project.artifactId}</sonar.moduleKey>
+    <sonar.projectKey>org.heigit.ohsome:oshdb</sonar.projectKey>
```
joker234, line 56: Does something like this work here?
```suggestion
    <sonar.projectKey>${project.groupId}:oshdb</sonar.projectKey>
```

joker234 (changes requested):

joker234:
Maybe additionally: Do we want to add the parent change to the Changelog?

pom.xml
tyrasd: let's try

tyrasd:
> Maybe additionally: Do we want to add the parent change to the Changelog?

IMHO such changes are not really something users need to know (and would search in a changelog document), except if it is a big/significant change which affects developers directly (like the change of the group id). but since we had a changelog entry for a parent upgrade already, I just generified that one a bit. :innocent:

joker234 (approved):

PR #387, opened 2021-06-17
Move oshdb-api's backend implementation into separate submodule(s)

tyrasd:
### Description

This creates a new submodules `oshdb-api-ignite` which contains the backend implementation code for running OSHDB queries on ignite.

I kept the JDBC/H2 backend implementation in the `oshdb-api` module for now, since it is a relatively lightweight (from a code and probably also from a dependencies point of view), and pulling it out would come with some annoyances: mainly that most tests (even the ones in ignite) use the H2 backend implementation as a "default" backend to run queries and/or to access keytables.

Some of the above points could be solved in the future (e.g. grid-included keytables, implementing a special in-memory "dummy" backend for tests, etc.). So, perhaps it's better to have a separate `oshdb-api-jdbc` (`oshdb-api-h2`) module already now? //edit: it would likely look more or less like this in at the moment: [#388](https://github.com/GIScience/oshdb/pull/388/files)

PS: This also upgrades ignite to the newest version (`2.10.0`).

### Corresponding issue

Closes #16 (?)
Closes #386 

### New or changed dependencies

- introduced new submodule `oshdb-api-ignite`
- moved ignite dependencies from `oshdb-api` to the new module

### Checklist
- [x] My code follows the [code-style](https://github.com/GIScience/oshdb/blob/master/CONTRIBUTING.md) rules, and I have checked on the [static analyses](https://jenkins.ohsome.org/job/oshdb/view/change-requests/) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) (if applicable) results
- [x] I have commented my code
- [x] I have written javadoc (required for public classes and methods)
- [x] I have added sufficient unit tests
- ~~I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/master/documentation)~~
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/oshdb/blob/master/CHANGELOG.md)
- ~~I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/issues/new) in the corresponding repository~~
[...]

PR #388, opened 2021-06-17
🚧 separate submodule for oshdb-api's JDBC/H2 implementation  🚧

tyrasd:
possible addedum to #387 to also add a `oshdb-api-jdbc` module.


tyrasd:
closing: we keep the "default" jdbc backend in the oshdb-api module.

PR #419, opened 2021-08-12
Some sonar-lint bugs/security hotspots

rtroilo:
this PR should, hopefully, fix the remaining bugs and security hotspots of sonar-lint

### Checklist
- ~I have commented my code~
- ~I have written javadoc (required for public classes and methods)~
- ~I have added sufficient unit tests~
- ~I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/issues/new) in the corresponding repository~
- ~I have adjusted the [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-benchmarks/-/issues/new) in the corresponding repository~


oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/db/OSHDBH2.java
```diff
@@ -80,29 +80,22 @@ public OSHDBH2 inMemory(boolean cache) throws SQLException {
...
-              "insert into " + tablename + "(" + columns + ") values (" + placeholders + ")"
-          );
-          rs = srcStmt.executeQuery("select " + columns + " from " + tablename);
-          while (rs.next()) {
-            for (int i = 1; i <= columnNames.size(); i++) {
-              destStmt.setObject(i, rs.getObject(i));
+          String insertSql = insertSql(tablename, columnNames, columns);
+          String querySql = String.format("select %s from %s", columns, tablename);
+          try (var destStmt = this.connection.prepareStatement(insertSql);
+              var rs = srcStmt.executeQuery(querySql);) {
```
joker234, line 89: Is `;` usually placed at the end of a `with` statement?
```suggestion
              var rs = srcStmt.executeQuery(querySql)) {
```

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/tagtranslator/TagTranslator.java
```diff
@@ -72,33 +78,30 @@ public TagTranslator(Connection conn) throws OSHDBKeytablesNotFoundException {
 
     // test connection for presence of actual "keytables" tables
     EnumSet<TableNames> keyTables =
-        EnumSet.of(TableNames.E_KEY, TableNames.E_KEYVALUE, TableNames.E_ROLE);
+        EnumSet.of(E_KEY, E_KEYVALUE, E_ROLE);
     for (TableNames table : keyTables) {
       try (Statement testTablePresentQuery = this.conn.createStatement()) {
-        testTablePresentQuery.execute("select 1 from " + table.toString() + " LIMIT 1");
+        var select = format("select 1 from %s limit 1", table);
+        testTablePresentQuery.execute(select);
```
joker234, line 85: In `OSHDBH2.java` you named all variables containing a String typed SQL statement `…Sql`. If you want to do something like that in general, you could use a name like this here:
```suggestion
        var selectSql = format("select 1 from %s limit 1", table);
        testTablePresentQuery.execute(selectSql);
```

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/backend/MapReducerJdbc.java
```diff
@@ -48,29 +46,52 @@ public boolean isActive() {
     return true;
   }
 
-  protected ResultSet getOshCellsRawDataFromDb(CellIdRange cellIdRange)
+  protected List<Map.Entry<String, CellId>> getOshTableCellIds(CellIdRange cellIdRange)
```
joker234, line 49: Do we have unittests for this function? If not, I think we should have some.

joker234 (commented):
ToDo:
- [ ] benchmark changes in ed1a850c

oshdb-etl/src/main/java/org/heigit/ohsome/oshdb/tool/importer/extract/Extract.java
```diff
@@ -294,20 +294,19 @@ public void sortByFrequency(KeyValueFrequencyCollector kvFrequency)
...
-              throw new RuntimeException(e);
-            }
-          }).count();
+      }, maxSize, KeyValuePointer::estimateSize)
+          .with(KeyValuePointer::write, KeyValuePointer::read)
+          .sort(keys.iterator());
+      while (sort.hasNext()) {
+        var kvp = sort.next();
+        kvp.write(keyValuesDataOutput);
+      }
```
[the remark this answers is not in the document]
rtroilo: the count in the first place was quite silly! as it's only reason was to consume the stream.

oshdb-etl/src/main/java/org/heigit/ohsome/oshdb/tool/importer/extract/Extract.java
```diff
@@ -321,20 +320,17 @@ public void sortByFrequency(RoleCollector roleFrequency)
...
         }
         return a.role.compareTo(b.role);
-      }, maxSize, Role::estimateSize).with(Role::write, Role::read).sort(roleFrequency.iterator()))
-          .peek(r -> {
-            try {
-              r.write(rolesDataOutput);
-            } catch (IOException e) {
-              throw new RuntimeException(e);
-            }
-          }).count();
```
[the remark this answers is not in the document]
rtroilo: see above answer.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/backend/MapReducerJdbc.java
```diff
@@ -48,29 +46,52 @@ public boolean isActive() {
...
   }
 
   /**
    * Returns data of one cell from the raw data stream.
    */
-  protected GridOSHEntity readOshCellRawData(ResultSet oshCellsRawData)
-      throws IOException, ClassNotFoundException, SQLException {
-    return (GridOSHEntity)
-        (new ObjectInputStream(oshCellsRawData.getBinaryStream(1))).readObject();
+  protected GridOSHEntity readOshCellRawData(Map.Entry<String, CellId> entry) {
```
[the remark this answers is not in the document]
rtroilo: this part is covered by tests, but you are right there is not a specific test case for this method/class only.
Should we add one in this PR?

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/backend/MapReducerJdbc.java
joker234: see #420

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/backend/MapReducerJdbc.java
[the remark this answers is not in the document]
joker234: we don't have to in this PR, see #420

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/time/TimestampFormatter.java
```diff
@@ -11,28 +11,8 @@
...
-   * @return a timestamp formatter object
-   */
-  public static TimestampFormatter getInstance() {
-    if (instance == null) {
-      instance = new TimestampFormatter();
-    }
-    return instance;
+  private TimestampFormatter() {
+    // utility class
   }
```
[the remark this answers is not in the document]
rtroilo: the background here was/is, that SimpleDateFormat is not thread safe, so using one per thread solves this. But TimestampFormatter was/is a singleton so removing those thread-local was not possible, 
I would like to see/know if this current solution has any negative impacted in performance, if not I think this is much cleaner.
Let's see the benchmarks for this.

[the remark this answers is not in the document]
rtroilo: an other solution would be to make TimestampFormater a singleton again but use synchonized-blocks to ensure, that the Dateformater is not concurrently used.
https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html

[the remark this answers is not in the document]
tyrasd: 1. AFAICS, the class is actually not really used in the OSHDB itself (apart from a single usage in `MapReducer.java` which would be "trivial" to reimplement). It is however used quite extensively by the ohsome API in data extraction requests.
2. For the usage in the ohsome API, the current implementation with threadlocal `SimpleDateFormat`ers is actually quite crucial for performance: we don't want to re-parse the data format pattern every time we want to stringify a datetime value. The singleton+threadlocal pattern was the easiest I could come up with.
3. So I guess we should either a) remove the class here and put the more complex but better performing logic into the ohsome API, or b) keep it as is.

does that make sense?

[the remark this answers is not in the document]
rtroilo: > 2\. For the usage in the ohsome API, the current implementation with threadlocal `SimpleDateFormat`ers is actually quite crucial for performance: we don't want to re-parse the data format pattern every time we want to stringify a datetime value. The singleton+threadlocal pattern was the easiest I could come up with.

Do  you have any numbers on here, would be interesting to know how negative the affect of creating new DateFormater actually is.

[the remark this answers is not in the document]
rtroilo: > 1. AFAICS, the class is actually not really used in the OSHDB itself (apart from a single usage in `MapReducer.java` which would be "trivial" to reimplement). It is however used quite extensively by the ohsome API in data extraction requests.

I looked through the reference to `TimestampFormatter` in the ohsome-api, and as far as I can tell only in the 
- [DataExtractionTransformer](https://github.com/GIScience/ohsome-api/blob/0965541e7a24171cd1d11017eed7fc781b29b851/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/DataExtractionTransformer.java#L144)
- [ElementsRequestExecutor](https://github.com/GIScience/ohsome-api/blob/0965541e7a24171cd1d11017eed7fc781b29b851/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java#L121)
the LocalThread mechanism is used. And also only for `isoDateTime`.

So my preference is than to remove the TimestampFormatter from the oshdb-api and move it to the ohsome-api maybe in a slightly modified version. For example a TimestampFormatter per request with proper clean up afterwards. this should then not influence performance.

What are your  thoughts about this? And should we already do it in this PR, if not, I will revert this change and we should create an issue.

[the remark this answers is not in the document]
tyrasd: > Do you have any numbers on here, would be interesting to know how negative the affect of creating new DateFormater actually is.

I created a simple [microbenchmark for this](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/microbenchmarks/timestampformatter-benchmark). These are the results:

```
Benchmark                          Mode  Cnt  Score   Error  Units
TimestampFormatterBench.master    thrpt    8  2.753 ± 0.522  10⁶ ops/s
TimestampFormatterBench.pr419     thrpt    8  0.919 ± 0.211  10⁶ ops/s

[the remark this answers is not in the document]
Comment (rtroilo) [reply oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/time/TimestampFormatter.java:16]: Thank you, and yes sure, there is a performance difference between creating always a new one. 
But I would be interested, if we can see this in the application/request. I guess this is more cumbersome to measure this.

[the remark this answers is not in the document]
Comment (tyrasd) [reply oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/time/TimestampFormatter.java:16]: I think it's pretty much neglectable in the actual application. Yes, it might get called many times in a data extraction request, and the overhead is significant (~3x as seen above). But for data extraction requests, there should be other parts of the processing taking much longer than this (e.g. the building of the geojson output for example), and the real bottleneck will practically always be the network.

I had thought that the overhead of recreating the formatter objects would be much larger. So, given that that's not the case, I guess we could simplify the implementation here. But then the question remains: why do we need to ship this thin wrapper around an already built-in java functionality? Woudn't it be better drop the class altogether?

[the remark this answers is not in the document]
Comment (rtroilo) [reply oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/time/TimestampFormatter.java:16]: Thank you and I totally agree with you. So dropping this class in the oshdb-api, replacing the functionality in the oshdb-api and find a new (or same) solution for the ohsome-api.
So I do the first 2 parts in the oshdb-api right now in this PR.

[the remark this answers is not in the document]
Comment (rtroilo) [reply oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/time/TimestampFormatter.java]: Done and added to the changelog.

File: CHANGELOG.md [hunk 8]
```diff
@@ -3,6 +3,11 @@ Changelog
 
 ## 0.8.0-SNAPSHOT (current master)
 
+### breaking changes
+ * removing of `oshdb-util:util.time.TimestampFormatter` ([#419])
```
tyrasd, line 7: usually we write changelog entries with active verbs, instead of passive (…ing) sentences.

```suggestion
 * remove `oshdb-util:util.time.TimestampFormatter` class ([#419])
```

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/db/OSHDBJdbc.java
```diff
@@ -80,19 +80,19 @@ public OSHDBJdbc prefix(String prefix) {
...
-      } else {
-        return null;
+      try (var result = stmt.executeQuery()) {
+        if (result.next()) {
+          return result.getString(1);
+        }
       }
     } catch (SQLException ignored) {
-      return null;
+      // ignored
```
tyrasd, line 93: What about throwing a `OSHDBTableNotFoundException` here instead of silently ignoring the error (which wold most likely happen because an OSHDB file doesn't have metadata table). Especially if we want to use this in the future to check for compatibility of a given OSHDB file, we need a valid metadata table anyway.

rtroilo: 👍🏽  makes sense, I update it.

rtroilo: What do you think about to return an empty string instead of null in case there is no properties that match.

tyrasd: hmm, not a big fan because the value stored in the table could actually be an empty string. But the method could return an empty `Optional<String>` if a value was absent.

rtroilo: > What about throwing a `OSHDBTableNotFoundException` here instead of silently ignoring the error (which wold most likely happen because an OSHDB file doesn't have metadata table). Especially if we want to use this in the future to check for compatibility of a given OSHDB file, we need a valid metadata table anyway.

It could also be the case that the database/connection in this moment, is not valid any more, so I throw better a OSHDBException and pass the SQLExcetption to it.

rtroilo: > hmm, not a big fan because the value stored in the table could actually be an empty string. But the method could return an empty `Optional<String>` if a value was absent.

fine for me, so I don't change this now.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/backend/MapReducerJdbc.java
```diff
@@ -79,40 +100,13 @@ protected GridOSHEntity readOshCellRawData(ResultSet oshCellsRawData)
...
-                  oshCellsRawData.close();
-                }
-                return data;
-              } catch (IOException | ClassNotFoundException | SQLException e) {
-                throw new RuntimeException(e);
-              }
-            }
-          }, 0
-      ), false);
+      return oshCellIds.stream().map(this::readOneGridCell);
```
tyrasd, line 107: can you explain what the rationale was behind this change? Previously this executed a single sql query to fetch all affected cell-id-ranges in a single swoop. Now it would fetch every cell individually in single sql queries, which could come with a (large?) performance overhead.

What is the benefit of doing it in this way?

rtroilo: The benefit is to close the resources properly, performance wise this should not introduce a large overhead, because for fetching a single cell the indexes could be fully used.
An other benefit is that cellIds is a full list of ids to process and in theory this stream could be very good parallelized. In theory only because for this to work properly `OSHDBJdbc` should not return a single share connection but from a pool of connection. But this refactoring we should do in an separate PR.
And for `h2` databases, smaller queries are better because `h2` does not have a  cursor and always builds up the full resultset (we can see this by increase file size) before returning.

tyrasd: ok, that sound legit.

however, this feels like a quite major change. and here it is presented in a pull request which claims to be only about fixing sonarlint issues, where I would assume that there are no changes to any algorithms or internal business logic. Maybe we can split this change off into a separate PR and take a more [in depth](https://github.com/GIScience/oshdb/pull/419#pullrequestreview-728792474) look there?

rtroilo: actually it is fixing this [sonar bugs](https://sonarcloud.io/project/issues?id=org.heigit.ohsome%3Aoshdb&issues=AXjvsOpdmEE7-fvh3PD5&open=AXjvsOpdmEE7-fvh3PD5)  by closing the PreparedStatment properly.
The changes around it just where the consequence of this.
But we could life this out and do it in a separate issue/pr.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ Changelog
 
 ## 0.8.0-SNAPSHOT (current master)
 
+### breaking changes
+ * remove class `oshdb-util:util.time.TimestampFormatter` ([#419])
+
+[#419]: https://github.com/GIScience/oshdb/pull/419
+
```
joker234, line 10:
```suggestion
### breaking changes

 * remove class `oshdb-util:util.time.TimestampFormatter` ([#419])

[#419]: https://github.com/GIScience/oshdb/pull/419

```

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/mapreducer/backend/MapReducerJdbc.java
tyrasd: >  we could leave this out and do it in a separate issue/pr.

:+1:

tyrasd: I've created #422 for this.

oshdb-api/src/main/java/org/heigit/ohsome/oshdb/api/db/OSHDBJdbc.java
tyrasd: follow up ticket: #421

tyrasd (approved):
lgtm

PR #426, opened 2021-09-08
fix serializability of geometry filters

tyrasd:
### Description
Fixes a serializability issue involving the geometry based filters (`length:` and `area:`).

This currently causes ohsome API queries to fail if they use these filters: [example](https://api.ohsome.org/v1/elements/count?bboxes=8.67%2C49.39%2C8.71%2C49.42&filter=type%3Away%20and%20natural%3D*%20and%20length%3A(0..10)&format=json&time=2014-01-01%2F2017-01-01%2FP1Y)

### Checklist
- [x] I have written javadoc (required for public classes and methods)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,12 @@ Changelog
 
  * remove class `oshdb-util:util.time.TimestampFormatter` ([#419])
 
+### bugfixes
+
+* fix a bug which causes queries using the geometry filters `length:` and `area:` to fail when executed on an ignite cluster ([#426])
+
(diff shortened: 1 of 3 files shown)


rtroilo (changes requested):
I think we should add a test, which could show us if there is a problem with serializability in our code.
Do we have a test that is capable to do so and or how much would be the effort to write such a test?

tyrasd:
>  Do we have a test that is capable to do so 

not at the moment

> and or how much would be the effort to write such a test?

not sure how one would approach it.

rtroilo (approved):

PR #438, opened 2021-12-22
never return negative values for areas

tyrasd:
Invalid polygons can potentially in some certain situation have a bigger inner rings area than the outer ring area. The area method should however never return a value smaller than zero. This clips it to a minimum value of zero in such (errror) cases.

### Checklist
- ~~I have commented my code~~
- ~~I have written javadoc (required for public classes and methods)~~

### DIFF
--- a/oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/Geo.java
+++ b/oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/Geo.java
@@ -164,7 +164,7 @@ public class Geo {
     for (int i = 0; i < poly.getNumInteriorRing(); i++) {
       area -= Math.abs(ringArea((LinearRing) poly.getInteriorRingN(i)));
     }
-    return area;
+    return Math.max(0, area);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/Geo.java
```diff
@@ -164,7 +164,7 @@ public static double areaOf(Polygon poly) {
     for (int i = 0; i < poly.getNumInteriorRing(); i++) {
       area -= Math.abs(ringArea((LinearRing) poly.getInteriorRingN(i)));
     }
-    return area;
+    return Math.max(0, area);
```
bonaparten, line 167: You could specify here or in the java doc of the method why you use the Math.max function or that the area cannot be < 0 . It is just a thought, you don't have to.

bonaparten (dismissed):

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/geometry/Geo.java
joker234: Good thought. Ideally this would be reflected in the [ohsome API docs](https://github.com/GIScience/ohsome-api/tree/master/docs) as well.

tyrasd: I've added a sentence about it to the javadoc. not sure if it really needs to be mentioned anywhere else, since I would thing that everybody would assume that areas can never be negative values.

joker234 (approved):

PR #476, opened 2022-11-10
some minor additions

rtroilo:
Some minor additions,
- more methods to create osm entities
- widening tagTranslator methods arguments

### Checklist
- ~[ ] I have commented my code~
- ~[ ] I have written javadoc (required for public classes and methods)~
- ~[ ] I have added sufficient unit tests~
- ~[ ] I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/master/documentation)~
- ~[ ] I have updated the [CHANGELOG.md](https://github.com/GIScience/oshdb/blob/master/CHANGELOG.md)~
- ~[ ] I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/issues/new) in the corresponding repository~
- ~[ ] I have adjusted the [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-benchmarks/-/issues/new) in the corresponding repository~

### DIFF
--- a/oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/tagtranslator/JdbcTagTranslator.java
+++ b/oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/tagtranslator/JdbcTagTranslator.java
@@ -6,6 +6,7 @@ import static org.heigit.ohsome.oshdb.util.tagtranslator.ClosableSqlArray.create
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.collect.Maps;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -82,7 +83,7 @@ public class JdbcTagTranslator implements TagTranslator {
   }
 
   @Override
-  public Map<OSMTag, OSHDBTag> getOSHDBTagOf(Set<? extends OSMTag> tags) {
+  public Map<OSMTag, OSHDBTag> getOSHDBTagOf(Collection<OSMTag> tags) {
     var keyTags = Maps.<String, Map<String, OSMTag>>newHashMapWithExpectedSize(tags.size());
     tags.forEach(tag -> keyTags.computeIfAbsent(tag.getKey(), x -> new HashMap<>())
         .put(tag.getValue(), tag));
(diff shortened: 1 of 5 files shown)


oshdb/src/main/java/org/heigit/ohsome/oshdb/osm/OSM.java
```diff
@@ -42,6 +55,21 @@ public static OSMRelation relation(final long id, final int version, final long
...
   }
 
+  public static OSMRelation relation(final long id, final int version, final long timestamp,
+      final long changeset, final int userId, final List<OSHDBTag> tags, final OSMMember[] members) {
+    return new Relation(id, version, timestamp, changeset, userId, toKV(tags), members);
+  }
+
+  private static int[] toKV(List<OSHDBTag> tags) {
+    var kv = new int[tags.size() * 2];
+    for (var i=0; i < tags.size(); i++) {
```
mmerdes, line 65: (if we had a list/collection i would prefer to iterate over the `tags`list in order to save the index computation)

oshdb-util/src/main/java/org/heigit/ohsome/oshdb/util/tagtranslator/JdbcTagTranslator.java
```diff
@@ -82,7 +83,7 @@ public Optional<OSHDBTag> getOSHDBTagOf(OSMTag tag) {
   }
 
   @Override
-  public Map<OSMTag, OSHDBTag> getOSHDBTagOf(Set<? extends OSMTag> tags) {
+  public Map<OSMTag, OSHDBTag> getOSHDBTagOf(Collection<OSMTag> tags) {
```
mmerdes, line 86: is the narrowing to concrete `OSMTag` instances intentional?

mmerdes (approved):
lgtm - see a few minor comments

oshdb/src/main/java/org/heigit/ohsome/oshdb/osm/OSM.java
rtroilo: As kv is twice the size of tags. We need some sort of index to calculate the k and v position
```
 kv[i * 2] = key
 kv[i * 2 + 1] = value
 ```
 
What do you think about this variant:
 
 ```
 var i = 0;
 for (var tag : tags) {
  kv[i++] = tag.getKey();
  kv[i++] = tag.getValue();
 }
 ```

mmerdes: i know that the result structure is twice as long, but as we access it only in ascending order it would be safe to just append the 2 new output entries (per input entry) to the result.
(if we had a `Collection` type instead of an array).
as the _array_ is also used in the consuming type we should probably keep it.

mmerdes: and yes, the second variant looks nicer to me
:)

mmerdes (approved):
looks good as discussed

joker234 (approved):
Technical approve as @mmerdes already approved the code

PR #503, opened 2023-04-17
Remove left over throw documentation

### LINKED ISSUE #498: TagTranslator returns Null if OSMTag is unknown
### Bug Description

`TagTranslator.lookupTag(OSHDBTag)` returns `Null` if the `OHSDBTag` cannot be found.

This is confusing because

- the documentation states that ` @throws OSHDBTagOrRoleNotFoundException if the given tag cannot be found` . There are only some functions that throw this exception.
- all `getOSHDB...` return an optional while all `lookup..` methods return an object. 

The behaviour of `TagTranslator` may need to be streamlined.

rtroilo:
There was a documentation about throwing `OSHDBTagOrRoleNotFoundException` when looking up a oshdb-tag, which is never thrown anymore!

### Corresponding issue
Closes #498

### New or changed dependencies
-

### Checklist
- [X] My code follows the [code-style](https://github.com/GIScience/oshdb/blob/master/CONTRIBUTING.md) rules, and I have checked on the [static analyses](https://jenkins.ohsome.org/job/oshdb/view/change-requests/) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) (if applicable) results
- [X] I have commented my code
- ~I have written javadoc (required for public classes and methods)~
- ~I have added sufficient unit tests~
- ~I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/master/documentation)~
- ~I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/issues/new) in the corresponding repository~
- ~I have adjusted the [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-benchmarks/-/issues/new) in the corresponding repository~


SlowMo24 (changes requested):
I think the documentation needs to be clearer what a NULL response means (see https://github.com/GIScience/oshdb/issues/498#issuecomment-1512853946)

tyrasd:
> I think the documentation needs to be clearer what a NULL response means (see [#498 (comment)](https://github.com/GIScience/oshdb/issues/498#issuecomment-1512853946))

This is a little bit of a tricky one. In my opinion it is sufficiently clear that the intended way to work with the `TagTranslator` is to call the `.lookup*` methods with `OSHDBTag` objects which have been previously returned by the same `TagTranslator` using one of the respective `.get*` methods. When constructing `OSHDBTag` objects manually (or mixing-and-matching different `TagTranslator`s using different keytable sources), the resulting behavior is basically just undefined. The result might be `null`, it might throw an exception, or do something unexpected altogether. I try to mention this in the javadoc of the `TagTranslator` interface.

SlowMo24 (approved):

PR #512, opened 2023-08-31
bugfix: areaOfInterest should not override previously set bbox

tyrasd:
instead, it should use the intersection of the previously set bbox and the "incoming" one, to be consistent with all "filters" work on a mapreducer.


rtroilo (approved):
thank you, lgtm

PR #495, opened 2023-03-23
CellIterator: add "lastContributionTimestamp" to OSMEntitySnapshot & decouple from Grid implementation

tyrasd:
### Description
Includes the following aspects of the `CellIterator` class:
* return the `lastContributionTimestamp` for each entity snapshot (i.e. the timestamp when the OSM entity was last touched before the snapshot timestamp)
* decouple `CellIterator` from grid implementation: It now accepts an `OSHEntitySource` (i.e. a stream of `OSHEntity` object plus the corresponding bounding box) instead of a raw grid cell object. This makes it easier to swap this out in the future with alternative backends that don't rely on this exact implementation of the grid
* refactor internal "data holding" classes to records

### Checklist
- [x] My code follows the [code-style](https://github.com/GIScience/oshdb/blob/master/CONTRIBUTING.md) rules, and I have checked on the [static analyses](https://jenkins.ohsome.org/job/oshdb/view/change-requests/) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) (if applicable) results
- [x] I have commented my code
- [x] I have written javadoc (required for public classes and methods)
- [x] I have added sufficient unit tests
- ~~I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/master/documentation)~~
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/oshdb/blob/master/CHANGELOG.md)
- ~~I have adjusted the [examples](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-examples/-/issues/new) in the corresponding repository~~
- ~~I have adjusted the [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/oshdb-benchmarks/-/issues/new) in the corresponding repository~~

PR #516, opened 2023-09-29
Fix performance degradation in ignite streaming endpoint

tyrasd:
#495 had unfortunately introduced a bug[^1] in _affinitycall_'s stream implementation which caused the backend to always use the slow fallback path (i.e. `.collect().stream()`). This is particularily slow for medium to large bounding boxes, as it will not use the preflight optimizations implemented for `stream`.

For example, this has the effect that with this version all ohsome-api data extractions beeing very slow.

[^1]: the concrete problem being that `stream.toList()` returns an immutable list which cannot be shuffled in the following code line, while the previously used `stream.collect(Collectors.toList())` does return a mutable list (even though it does not guarantee the mutability of it, btw).

### Checklist
- [x] I have written javadoc (required for public classes and methods)


rtroilo (approved):
thank you!

PR #525, opened 2024-02-22
Jenkins: use default "test-compile" setting for reports and statistics

tyrasd:
related to #518


ConstantinNicolai (approved):
looks correct

PR #529, opened 2024-06-07
fix crash while building geometry of multipolygon with zero-node way members

tyrasd:
this bug occurs when the following two conditions are met:

* a multipolygon refers to a member which at the given timestamp has zero nodes
  - example: https://www.openstreetmap.org/api/0.6/relation/22050/history at `2008-09-30T10:00:52Z` contains https://www.openstreetmap.org/api/0.6/way/25714579/6 which has zero nodes
  - AFAIK only old versions of the OSM API allowed such "invalid" ways to exists. Now there is a [check](https://github.com/openstreetmap/openstreetmap-website/blob/349200f/app/models/way.rb#L167) in [place](https://github.com/openstreetmap/cgimap/blob/8bfb626/include/cgimap/api06/changeset_upload/way.hpp#L56) which prevents these corner cases
* the order of the relation members is such that the 0-node member's previous members do not form a closed loop

### Checklist
- [x] My code follows the [code-style](https://github.com/GIScience/oshdb/blob/main/CONTRIBUTING.md) rules, and I have checked on the [static analyses](https://jenkins.ohsome.org/job/oshdb/view/change-requests/) and [benchmark](https://reports.ohsome.org/oshdb-benchmarks/) (if applicable) results
- ~~I have written javadoc (required for public classes and methods)~~
- ~~I have made corresponding changes to the [documentation](https://github.com/GIScience/oshdb/tree/main/documentation)~~
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/oshdb/blob/main/CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Changelog
 
 ## 1.3.0-SNAPSHOT (current main)
 
+### bugfixes
+
+* Fix crash while building geometry of multipolygon with zero-node way members under certain circumstances
+
 
 ## 1.2.2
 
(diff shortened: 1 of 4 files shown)


rtroilo (approved):
💯

